PR #13, opened 2026-01-06
Hide battery indicator on macOS builds

j0uni:
The battery indicator shows 0% on macOS builds, which is incorrect. This change hides the battery indicator on macOS to avoid displaying misleading information while maintaining the layout structure.

- BatteryIndicator widget returns SizedBox.shrink() on macOS
- Device screen battery indicator hidden on macOS
- Settings screen battery info row conditionally rendered (excluded on macOS)

### DIFF
--- a/lib/screens/device_screen.dart
+++ b/lib/screens/device_screen.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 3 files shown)


zjs81:
Do you know why its displaying the wrong information maybe we can fix it instead of hide it?

j0uni:
Yes, that would be correct approach. Closing this one.

PR #17, opened 2026-01-07
Fixed icons not being visible in Dark mode

wel97459:
Changed which color is applied to the icons. also update settings icons as well.

Screenshot_20260107_011350
Screenshot_20260107_011518


zjs81:
Looks good

PR #24, opened 2026-01-10
Added support for room server chat

wel97459:
Added login interface for room servers.
Added message support for messages pushed from room server.
Fixed notification to format room server messages.


zjs81:
Very Nice I'm reviewing it.

PR #34, opened 2026-01-14
fixing some german translations.

ericszimmermann:
fixed some german translation.

I do not have access to all windows, therefore some context may be wrong.
Decided to translate:
- "Anounce" to "Ankündigung"
- "Flood" zu "Fluten"
- Repeater has been reverted to the DEnglisch Repeater, because the only german alternative "Verstärker" (amplifier) does not transport the right meaning.

Sincerely Eric


ericszimmermann:
Sorry there is an other change, which I have unfortunately commited on the same branch:
To disable the rotation of the map.

zjs81:
@ericszimmermann  Thanks for the PR would you be able to clean up this PR so its just your translation changes?

ericszimmermann:
See new Pull-request

PR #38, opened 2026-01-15
Added public key in contacts list and in the repeater hub

wel97459:
Added public key in contacts list and in the repeater hub much like the original app:
Screenshot_20260114_233058
Screenshot_20260114_194317


zjs81:
Thank you, I will review this later today.

zjs81:
nice job

PR #41, opened 2026-01-16
Show repeater login error in login dialog

mtlynch:
For the repeater login dialog, if the login fails, the error is bubbled up as an Exception and displayed as a temporary snackbar notice behind the foreground dialog, making it difficult for the user to see. Also, the fact that it's an Exception means that it doesn't get localized to other languages other than English.

This change turns the login failure Exception into standard text in the login dialog and localizes it to support every language.

### Before

flutter_03

### After

flutter_01

### Note

I used gpt-5.2-codex to generate the translations. I only wrote the English text by hand and have not verified the text in other languages.


ericszimmermann:
German translation of the changes are correct. 👍

zjs81:
Nice job just merged

PR #44, opened 2026-01-17
Add a Github Action to build code in CI

mtlynch:
This adds a CI workflow in Github Actions to verify that the flutter builds compile for all supported platforms. This is helpful for reviewing external contributions, as it tells contributors immediately if their PR caused a compilation error.

I tried adding Windows, but it currently fails, so I excluded it from this initial set.


zjs81:
Just merged thank you

PR #46, opened 2026-01-17
German translation V2

ericszimmermann:
Reworked german translations on current version for clean Pull-reequest.

I do not have access to all windows, therefore some context may be wrong.
Decided to translate:

    "Anounce" to "Ankündigung"
    "Flood" zu "Fluten"
    Repeater has been reverted to the DEnglisch Repeater, because the only german alternative "Verstärker" (amplifier) does not transport the right meaning.

Sincerely Eric


zjs81:
Thank you I will merge it in.

PR #47, opened 2026-01-17
Disable Map rotation

ericszimmermann:
The map rotation is not that comfortable on Touch-devices.
Therefore I think this code should prevent rotation of the map.
Sincerely Eric


zjs81:
Thank you I'll merge it

PR #48, opened 2026-01-18
Added GPS enable and GPS interval settings.

wel97459:
There is a few things that need to be looked into, like how we know if a companion has a GPS.
And we should probable save the interval.
Requested by dragonidium95


Danie10:
Just wanted to ask about this feature (I did not see any sort of desription in the readme) when it says in the Location option, Interval for GPS in seconds, is that only for the frequency of the GPS to update the lattitude and logitude inside the app? It does not actually do any flood advert of the location?

I'm asking as I know the official app has no auto location sending at all. And I'm really looking for a toggle for sending a  location every 15 mins or so while mobile. If it does not do this, not sure if my logging an issue for it could be considered? I know the aim is to keep network traffic low.

PR #52, opened 2026-01-19
Added support for lm-studio

wel97459:
Added support for lm-studio


446564:
@wel97459 please checkout conflicts and then I can test it out.

Issue #54, opened 2026-01-20
Accessability: a few unlabeled buttons when using talkback


shortwavesurfer2009:
When on the channel's screen, I've found that the "+" icon in the bottom right hand corner reads as "unlabeled" and so does the filter button at the bottom right of the map tab.

Edit: Oh yeah, I forgot about the send message button. Went in a channel.

Edit 2: The information button in a chat as well.

<img width="814" height="1808" alt="Image" src="https://github.com/user-attachments/assets/a949687f-4ed6-49af-ba76-d16eecabdd9e" />

446564:
Remaining:
- [x] send message in both channel and direct
<strike>- [ ] channel info button</strike> moved to context menu, will update that

PR #60, opened 2026-01-21
update tooltips

446564:
add missing tooltip:
- channels, add channel button
- map, filter nodes button


446564:
only part of #54, send buttons needs new localization property. Also applies to send button in chat screen fro direct message.

PR #69, opened 2026-01-21
More french translation updates

spfmoby:
Nothing special, just more french updates


spfmoby:
I found spanish in some translations, the IA that dit the job must have lost context at some point :)

zjs81:
Thank you I will review this shortly

PR #111, opened 2026-01-29
Added disconnection handling, and fixed state changing of navigation on connection.

wel97459:
Added error catching to requestBatteryStatus to call _handleDisconnection when it fails update.
Updated ScannerScreen to manage navigation state logic on connection.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -962,9 +962,8 @@ class MeshCoreConnector extends ChangeNotifier {
     try {
       await sendFrame(buildGetBattAndStorageFrame());
     } catch (e) {
-      // Reset flag on error to allow retry
-      // Don't disconnect on battery request failure - it may be transient
-      _batteryRequested = false;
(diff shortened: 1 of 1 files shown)


lib/connector/meshcore_connector.dart
```diff
@@ -959,7 +959,14 @@ class MeshCoreConnector extends ChangeNotifier {
     if (!isConnected) return;
     if (_batteryRequested && !force) return;
     _batteryRequested = true;
-    await sendFrame(buildGetBattAndStorageFrame());
+    try {
+      await sendFrame(buildGetBattAndStorageFrame());
+    } catch (e) {
+      // Reset flag on error to allow retry
+      print('Battery status request failed: $e');
```
zjs81: Remove Print Statement please

zjs81 (changes requested):

lib/connector/meshcore_connector.dart
wel97459: I removed it.

zjs81:
@cursor review

PR #117, opened 2026-01-31
Prevent disconnection handling when already disconnected

wel97459:
This cures a race condition that was messing up the disconnection handler.
Before the bluetooth device was fully connected _handleDisconnection() was being call from the lisener.


wel97459:
cursor review

446564:
LGTM

PR #129, opened 2026-02-05
remove direct msg notification prefix

### LINKED ISSUE #128: [UX] Remove 'new message from' in direct notification
The new message from prefix seems unnecessary and takes up much of the line as with #61

446564:
The prefix "New message from " takes up a lot of space and was not localized anyway.

closes #128

### DIFF
--- a/lib/services/notification_service.dart
+++ b/lib/services/notification_service.dart
@@ -118,7 +118,7 @@ class NotificationService {
 
     await _notifications.show(
       contactId?.hashCode ?? 0,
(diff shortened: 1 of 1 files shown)


ericszimmermann:
Thank you, I find it really good solution:
1 Notification with the latest message, updating so that my smartwatch/Gadgetbridge can read them.
And one notification, which can be filtered out Gadgetbridge, but has the info of messages count.
I like it.
Sincerely Eric

PR #132, opened 2026-02-05
add flutter test to actions

### LINKED ISSUE #124: Run action for tests
There is only one test so far, but we should add a step to flutter and dart analysis to test the ones we have.

<strike>We could also merge #123 into the flutter action and reorder so it checks format, analysis and then tests in one action flow.</strike>

446564:
closes #124


446564:
I'm just going to merge this as there is inly one test at the moment anwyay

PR #141, opened 2026-02-08
Implement PathTraceMapScreen and refactor path tracing functionality

wel97459:
This replaces the path tracing UI, with a map to show the path. a long with a list of the hops and there signal reports


446564:
I think my last ask would be for single repeater ping have the window just show both there and back. But that ca be later as it's still alpha

PR #144, opened 2026-02-09
add support for whipseros

### LINKED ISSUE #137: Unable to connect to a device with WhisperOS firmware.
An app is not discovering devices flashed with [WhisperOS firmware](https://ssaprus.works/).

446564:
needed a new ble prefix

Closes #137


446564:
I see no harm in going ahead and easy to remove if not needed.

PR #152, opened 2026-02-11
remove wakelock

### LINKED ISSUE #149: Android screen keeps on while the app is in the foreground
The app doesn't allow for the screen to go off according to the system settings while in foreground, keeping it on at full brightness indefinitely. I'm observing this behaviour in every app screen, be it menu or channel or whatever.
Meshcore Open v 5.0.0, Android 12, MIUI 13.0.10.

446564:
was being used to keep ble active which is not what it does

in early testing the ble remains connected with display off and also when switching apps

closes #149


446564:
I will look at adding a proper ble background service

446564:
So I've had this running all night without trouble, even going in and out of range with successful reconnects. I'm going to merge and look into whether we need a BG ble service or not to help iOS.

terminalvelocity23:
Thank you, I'll be waiting a release with this included :)

PR #158, opened 2026-02-11
Update .gitignore to exclude .gradle/

ChaoticLeah:
Every time I build I get a bunch of files under .grade/ so I added it to the gitignore


446564:
Just moved into the android section

PR #178, opened 2026-02-17
Correct Chinese translation

MGJ520:
Correct Chinese translation


MGJ520:
> Correct Chinese translation

PR #179, opened 2026-02-17
Correct Chinese translation

MGJ520:
Correct Chinese translation


MGJ520:
Resolved

zjs81:
Looks good only thing is you need to run the flutter gen command for translations so it generates the source files

MGJ520:
> Looks good only thing is you need to run the flutter gen command for translations so it generates the source files

Hello, my computer currently doesn't have these environments. Could you help me out?

zjs81:
Thanks for your work. I just made some fixes and I will get it merged.

PR #182, opened 2026-02-18
Handle RESP_CODE_ERR frames explicitly in connector

Specter242:
## Problem
`RESP_CODE_ERR` (protocol response code `1`) was falling through to the generic "unknown frame" path in the connector, which loses structured firmware error detail.

## Solution
- Add an explicit `respCodeErr` branch in `_processFrame`
- Parse optional secondary error byte when present
- Log firmware error responses via `_appDebugLogService?.warn(..., tag: 'Protocol')`

## Scope
This PR is intentionally limited to connector error-frame handling in:
- `lib/connector/meshcore_connector.dart`

Not included in this PR:
- protocol/version capability getters
- scope messaging UI
- charging indicator
- dependency/package updates

## Validation
- `flutter analyze` passes
- `flutter test` passes

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -1845,6 +1845,7 @@ class MeshCoreConnector extends ChangeNotifier {
     if (frame.length >= 81) {
       _clientRepeat = frame[80] != 0;
     }
+
     // Firmware reports MAX_CONTACTS / 2 for v3+ device info.
     final reportedContacts = frame[2];
     final reportedChannels = frame[3];


lib/connector/meshcore_connector.dart
```diff
@@ -105,6 +105,8 @@ class MeshCoreConnector extends ChangeNotifier {
   static const int _defaultMaxChannels = 8;
   int _maxContacts = _defaultMaxContacts;
   int _maxChannels = _defaultMaxChannels;
+  int? _deviceProtocolVersion;
+  String? _activeFloodScopeTag;
```
446564: This is never used anywhere.

Why was it added?

lib/connector/meshcore_connector.dart
```diff
@@ -1827,12 +1845,12 @@ class MeshCoreConnector extends ChangeNotifier {
   void _handleDeviceInfo(Uint8List frame) {
     if (frame.length < 4) return;
     _firmwareVerCode = frame[1];
+    _deviceProtocolVersion = frame[1];
```
446564: Can you explain the difference between `_firmwareVerCode` and `_deviceProtocolVersion` which are both being assigned the same value?

Specter242: You are right that they were redundant here. Updated in `8a804a3` to use a single source of truth: `deviceProtocolVersion` now aliases `_firmwareVerCode`, and `supportsFloodScope` also derives from `_firmwareVerCode`.

lib/connector/meshcore_connector.dart
Specter242: Good catch. Removed `_activeFloodScopeTag` in commit `8a804a3` since it was unused in this focused PR slice.

lib/connector/meshcore_connector.dart
446564: I don't think you need an alias as the variable `_firmwareVerCode` already served the purpose.

lib/connector/meshcore_connector.dart
```diff
@@ -208,6 +208,8 @@ class MeshCoreConnector extends ChangeNotifier {
   int? get batteryMillivolts => _batteryMillivolts;
   int get maxContacts => _maxContacts;
   int get maxChannels => _maxChannels;
+  int? get deviceProtocolVersion => _firmwareVerCode;
```
446564: Can you tell me where this is used or why we need it?

lib/connector/meshcore_connector.dart
```diff
@@ -208,6 +208,8 @@ class MeshCoreConnector extends ChangeNotifier {
...
+  bool get supportsFloodScope => (_firmwareVerCode ?? 0) >= 8;
```
446564: Can you explain the purpose of this getter?

lib/connector/meshcore_connector.dart
```diff
@@ -1753,11 +1755,23 @@ class MeshCoreConnector extends ChangeNotifier {
         break;
       case respCodeCustomVars:
         _handleCustomVars(frame);
+        break;
+      case respCodeErr:
```
446564, line 1758: Can you explain how this is an error code and how it differs from the default case below?

446564 (changes requested):

lib/connector/meshcore_connector.dart
Specter242: Good callout. `respCodeErr` is a defined protocol response (`respCodeErr = 1` in `meshcore_protocol.dart`), while `default` handles truly unknown/unmapped codes. I narrowed this PR to only explicit `RESP_CODE_ERR` handling and logging (`dc193be`, `50ab46e`).

lib/connector/meshcore_connector.dart
Specter242: Removed in follow-up commits (`dc193be`, `50ab46e`). `supportsFloodScope` is no longer included in this PR.

lib/connector/meshcore_connector.dart
Specter242: Removed in follow-up commits (`dc193be`, `50ab46e`). `deviceProtocolVersion` is no longer part of this PR to keep scope strictly incremental.

Specter242:
Follow-up update based on review:

I narrowed this PR to one incremental change only: explicit RESP_CODE_ERR handling in MeshCoreConnector with protocol-tagged logging.

Removed from this PR:
- deviceProtocolVersion getter
- supportsFloodScope getter
- any protocol-version aliasing/placeholder fields

Current diff vs main is limited to lib/connector/meshcore_connector.dart and focused on error-frame routing/logging.

Validation rerun after narrowing:
- flutter analyze passes
- flutter test passes

zjs81 (approved):
LGTM

PR #195, opened 2026-02-19
add rbenv support

benallfree:
Hello, just trying out this awesome effort :)

Not sure if anyone else has run into this, but at least on my system I'm running multiple versions of Ruby and it's helpful to have a `.ruby-version` because `flutter run` blew up with a CocoaPods error without it.


446564:
I don't think we're using any ruby so it might be unrelated

446564:
Ah macos development?

benallfree:
Yep exactly, thank you!

PR #201, opened 2026-02-20
fix(channels): make edit/delete actions use parent context after bottom sheet closes Supports Windows Animations Off In Android

just-stuff-tm:
Bottom sheet closes

adds support for android windows animations off

Root cause: edit/delete dialogs were opened from the sheet context after Navigator.pop, so context.mounted was false and follow-up actions never ran.

Also keeps async await/error handling for channel edit/delete so failures surface to users instead of silently dropping.

### DIFF
--- a/lib/screens/channels_screen.dart
+++ b/lib/screens/channels_screen.dart
@@ -522,7 +522,7 @@ class _ChannelsScreenState extends State<ChannelsScreen>
                     : context.l10n.channels_muteChannel,
               ),
               onTap: () async {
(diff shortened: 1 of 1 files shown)


just-stuff-tm:
all updated

just-stuff-tm:
@codex Review

just-stuff-tm:
@codex review

just-stuff-tm:
@codex review

just-stuff-tm:
@gemini-code-assist /review

PR #205, opened 2026-02-21
fix(appbar): prevent title overflow on narrow widths

just-stuff-tm:
Use width-aware layout in AppBarTitle to avoid RenderFlex overflows under tight title constraints and larger text scaling. Hide subtitle and signal indicators progressively when space is limited while preserving normal behavior on wider layouts.


wel97459:
Awesome i'll test it out

PR #210, opened 2026-02-22
Better french translations

spfmoby:
More translations.

I noticed some words are missing, for example when a new repeater or contact is discovered, the notification is (in french):
Nouveau Repeater découvert
Nouveau Chat découvert

I looked in the translation file and the translation source is "New {contactType} discovered"
but contactType is not translatable.


zjs81:
Thank you!

PR #212, opened 2026-02-22
chore: add PlatformInfo helper util

benallfree:
This PR introduces a `PlatformInfo` helper to query platform capabilities from one central place.

I noticed that the platform detection logic was a little spread out and it was becoming messy as I was working on Chrome compatibility.

In particular, a lot of the `Platform.*` stuff is not available on Chrome so things needed to get wrapped in a `kIsWeb` everywhere. 

Beyond browser compat, this allows us to improve a couple abstractions like `isDesktop`, etc.


benallfree:
Closing this one too since it's only needed for Chrome compat.

PR #218, opened 2026-02-23
Show overlaps in public keys of repeaters

wel97459:
Show overlaps in public keys of repeaters on the map.
Added a function to handle filtering markers placed on the map.


zjs81:
@codex review

PR #217, opened 2026-02-22
enh: Chrome compatibility

benallfree:
This PR should give us a fully working Chrome web client.

Live preview: https://meshcore.meshenvy.org

Review status:

|Platform | Ok? |
|----|-----|
| OSX | ✅ |
| Chrome | ✅ |
| iOS | ❓  |
| Android | ❓  |
| WIndows | ❓  |
| Linux | ❓  |

PR #223, opened 2026-02-23
feat: image sideloading

benallfree:
PR stack:

- #262 

This is kind of a crazy idea.

I made an open source asset server powered by Cloudflare workers called Mesh Asset Server https://github.com/MeshEnvy/mas

It allows uploading images that will be saved for 30 days. You get a 6-char hash back.

This allows us to upload images and display them in chat. If someone happens to be offgrid, they'll just see something like `i:f84jgf4`

https://github.com/user-attachments/assets/39114c21-a303-4184-beb4-b5d091fbcfd9

To be merged first:

- [x] #217


zjs81:
I like the idea however it seems insecure. Do you think you could encrypt the image with aes using the shared secret for the chat and or the secret key for the channel ?

benallfree:
@zjs81 oh snap encryption is a good idea. Then it becomes more of a general 'asset snippet' hosting, all the processing would need to be done client-side. Let me think more about that because encrypted assets per channel would be a pretty cool idea.

benallfree:
Ok this branch is updated now too.

This still seems like a cool idea to me but I'm having trouble with Chrome because it has to download the encrypted file versions to memory, decrypt them, and then display them. For some reason it's very slow in Chrome but not on MacOS.

ericszimmermann:
To not have two buttons at the bottom (gif and image), which take much space maybe we could keep one "Action" Button, with a pop-up for gifs and image loading. Maybe there will be even more options in feature.
(With that I second the Issue #257, however not to hide the Button, but to keep the screen more streamlined)

Best Regards
Eric

benallfree:
@ericszimmermann Yes I agree, kind of like the iphone interface where you tap on it to get a variety of choices: photos, video, camera, etc.

zjs81:
You could download it as a stream object and decrypt it in memory not sure if that would be any faster though.

446564:
after you get the merge conflicts sorted can you squash everything into one commit and push for review and testing?

446564:
closing as stale, please resubmit when ready noting the previous changes.

benallfree:
No problem. Just for clarity on this project, you don't squash merge?

446564:
> No problem. Just for clarity on this project, you don't squash merge?

we haven't been, but that's probably a better idea. or at least an option.

PR #229, opened 2026-02-24
feat: message splitting

benallfree:
Prerequisites:

- [ ] #217 

-------

This PR introduces a new 2000 char limit and will split the messages before sending out. I'm not sure if I love the ux yet. 

image


zjs81:
Whats the strat for preventing this from ddosing the mesh? This also doesnt guarantee delivery so I'm not sure I understand the utilty of this. Can you please answer the following?

1. What are the Pros of this. 
2. What are the cons of this. Assume that everyone switches from the regular meshcore app to this open app. 
3. It looks like you have multiple changes on this PR. Would you be able to split everything up and do 1 pr per feature so its easy to review?


Thank you I appreciate your work on this. 

Regards,

benallfree:
I'm not super convinced it's a good idea yet either. I hear people ask for the ability to write longer messages, but reassembling them on the other end seems error prone.

DDoS probably isn't a big concern in my mind because people are already writing bots with existing tools. The client could restrict to 2000 chars or something like that just to stop easy abuse. 

I'll clean up this branch, I'm sure it's the Chrome compat mixed in.

zjs81:
How about instead of doing this, you can use smaz, which is a text compression engine that's already built into the app. And measure the size of the message in bytes. And if they have smaz enabled, allow them to send longer messages. I feel like that would be better, and it would stay up to spec with the protocol.

benallfree:
Ok branch updated, much cleaner PR now. 

So *if* we want this, my idea was to do a rotating message identifier `a-z` so you could differentiate between the same chunk number meant for different messages.

```
[1/2a]This is part 1 of message A
[1/2b]This is part 1 of message B
[2/2a]This is part 2 of message A
[2/2b]This is part 2 of message B
```

Message chunks are also discriminated by node ID so as long as a single node doesn't flood the mesh with 26 different messages, all the clients should be able to reassemble them after all the pieces arrive with a reasonable timeout where it will give up waiting for missing pieces after maybe 30s or something.

But I don't know how this would affect messages in rooms or any other situation where messages are delayed for a long time, and that's where I started backing off from this idea altogether.

benallfree:
@zjs81 I did a little more research into smaz and I don't see any references to it in the MeshCore C++ firmware. Is it a Flutter-only feature? If so, it seems that anyone using a stand-alone device would not be able to communicate.

Either way, I agree that message splitting is not a great idea so I'm closing this PR.

zjs81:
Its a flutter widget I ported over

zjs81:
https://github.com/zjs81/meshcore-open/blob/main/lib/helpers/smaz.dart

PR #230, opened 2026-02-24
feat: Lua scripting

benallfree:
Prerequistes:

- [ ] #217 

Preview: https://meshcore.meshenvy.org/

This is more of a heads up, I'm not even sure you want it in here. It's part of my larger goal at MeshEnvy to introduce a programmable abstraction layer for the mesh. The idea is that users/devs would be able to develop small mesh apps that run either directly on-device or through companion client.

As an example, this could be used to make this companion client talk to a local LLM that is trained with local information, survival, first aid, rescue, etc. You could ask the node questions and it would answer.

I plan to implement some UI methods in the Mesh Abstraction Layer (MAL) as well, so these scripts could eventually have a basic UI associated with them.

If you like this idea or want to see how it evolves, I'll keep this updated.

IMHO allowing users to create (and install) scripts directly in the companion client would make things highly programmable.

https://github.com/user-attachments/assets/61203967-c5e6-4bb3-b66c-9d3e9f3c973b

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -12,6 +12,7 @@ include: package:flutter_lints/flutter.yaml
 analyzer:
   exclude:
     - vendor/lua_dardo/**
+    - vendor/darth_lua/**
 
 linter:
   # The lint rules applied to this project can be customized in the


zjs81:
This is cool. Have you tested it on a release build to see if it still works?

benallfree:
@zjs81 still very WIP, no real testing yet. I'll keep this PR updated though

zjs81:
I have not reviewed the code too much but my main concern is that on release build flutter does a tree collapse meaning it gets rid of anything that isnt used. So I'm not sure if this would work.

benallfree:
Local LLM has entered the chat

image

benallfree:
> I have not reviewed the code too much but my main concern is that on release build flutter does a tree collapse meaning it gets rid of anything that isnt used. So I'm not sure if this would work.

@zjs81 I'm not sure if I follow about what isn't being used that could get tree shaken but I'll try a prod release just to see anyway.

benallfree:
https://github.com/user-attachments/assets/62aab93a-81d7-4bca-be45-6efbc69d3fc0

benallfree:
This brings the chat companion pretty far from shore and turns it into an IDE.

If I continue this project, it makes more sense for it to be a separate app. However, I'm exploring moving scripting into the firmware so you don't need an app at all.

Closing.

PR #173, opened 2026-02-14
Add contact favorites filter with companion flags bit0 persistence

Specter242:
## Summary
- Adds a Favorites-only filter in contacts UI.
- Uses companion contact `flags` bit0 as the source of truth for favorites (read + write), instead of local app-only storage.
- Preserves contact flags when sending `CMD_ADD_UPDATE_CONTACT` path updates so favorites are not cleared by path changes.
- Addresses reviewer feedback about cross-client persistence by removing the app-only favorites storage path.

## Context
- Resolves #40
- Resolves #166
- Follow-up to #172 (this branch currently includes #172 commits and should be reviewed/merged after #172).

## Test plan
- [x] `dart format` on modified Dart files
- [x] `flutter gen-l10n`
- [x] `flutter analyze`
- [x] Build/install on Android (`flutter build apk --debug` + `adb install -r`)
- [ ] iOS test (not run: iOS device/environment unavailable)

Issue #166, opened 2026-02-12
Loosing Favorite flag


ericszimmermann:
Hi, this is related to my issue #40.

When I write a DM to a contact which is favored, or manage a repeater which was in my favorite List in Liams App here in meshcore open, then it looses this flag.
Since currently I have to switch between the Apps, this means the contact/repeater disappears from the List in Liams App, which is quite annoying.

Sincerely Eric

446564:
I'm assuming it's one of the reserved for future use flags.

https://github.com/meshcore-dev/MeshCore/blob/246d681b2bb9201952bf93d7970bce054ceb3a00/docs/payloads.md?plain=1#L47-L59

But will need to do some digging or ask Liam. But I suppose of we just read and wrote back the flags it doesn't matter what they are.

446564:
Liam to the rescue, huge thanks.

Contact flags:
https://github.com/meshcore-dev/MeshCore/blob/246d681b2bb9201952bf93d7970bce054ceb3a00/src/helpers/ContactInfo.h#L10

Favorite bit flag:
https://github.com/meshcore-dev/MeshCore/blob/246d681b2bb9201952bf93d7970bce054ceb3a00/src/helpers/BaseChatMesh.cpp#L78

And a very generous snippet from his app code:
```
static bool isFavourite(int contactFlags) {
    // this first bit is favourites flag
    return contactFlags & 0x01 == 0x01;
  }

  static int getPermissions(int contactFlags) {
    // this first bit is favourites flag, so we will skip that
    return contactFlags >> 1;
  }

  static bool hasPermissionTelemetryModeBase(int contactFlags) {
    return getPermissions(contactFlags) & Constants.telemetryPermissions.base == Constants.telemetryPermissions.base;
  }

  static bool hasPermissionTelemetryModeLocation(int contactFlags) {
    return getPermissions(contactFlags) & Constants.telemetryPermissions.location == Constants.telemetryPermissions.location;
  }

  static bool hasPermissionTelemetryModeEnvironment(int contactFlags) {
    return getPermissions(contactFlags) & Constants.telemetryPermissions.environment == Constants.telemetryPermissions.environment;
  }
```

zjs81:
This is done.

Issue #40, opened 2026-01-16
Contact list filter "Favorites"


ericszimmermann:
Implementation of the Favorite-Filter in the contact list.
It would help to have the relevant contacts handy.
(I would like to have it filtered, you could also pin them on the top of the list.)

The Favoritestatus is saved as bit0 in the flags of the contacts.

Sincerely Eric

ericszimmermann:
Related Issue is, when I access a Repeater to manage, the flags will be resettet.
This results in removing the repeater from the Favorite list in the Meshcore App from Liam.
We should realy preserve this flag and maybe use it ourself.
It is the first bit in the contacts flag.

ericszimmermann:
I want to pitch the Favorite function again.
You should be able to set/unset it and filter after it.
Because you can configure your companion to automatically add (only) User contacts on Discovery.
Since we have no Discovery screen to manually add contacts, I have to use this option.
But with this option commes another one: "Clear oldest contacts, when contactlist is full, except the ones who are marked favorites."
The favorite ones have the lowest bit set in the contact flags.

So I either have to go to Liams App to get the Discovery Screen to add Manually or
I have to get to his App to set the Favorites, so that some of my private contacts are staying permanently (I want to keep the History and Settings, even if I do not use these contacts regularly).

Sincerely Eric

zjs81:
This is implemented.

PR #234, opened 2026-02-24
favorites handling only

ericszimmermann:
This is a pull request, which **only** takes the changes for **favorite handling from PR #173 (Add contact favorites filter with companion flags bit0 persistence)**.
Thanks to Specter242 for these changes. He has done all the work.
I just **updated** it **to** the status of the current **zjs81/main**,
without the changes of #172 (Add room sync controls with polished room server status UI), which also ended up in PR #173.
Solves Issues #40 and #166.
I had it tested, that it works with my T114 companion and in combination with Liam's App.

Sincerely Eric


zjs81:
@Specter242 @ericszimmermann Thanks for your contributions on this I will merge it in.

Issue #237, opened 2026-02-25
Add advanced repeater settings controls in app UI


Specter242:
## Summary
Expose advanced repeater controls in the repeater settings screen and wire refresh actions for those settings.

## Context
Current repeater settings UI does not expose all advanced controls that are available/expected for power users.

## Proposed Scope
- Add an "Advanced settings" section/card in repeater settings UI
- Include controls for:
  - interference threshold
  - AGC reset interval
  - flood max hops
  - multi-ACKs
- Add refresh actions for advanced fields
- Add l10n keys for new labels/actions

## Acceptance Criteria
- Advanced controls are visible and editable in repeater settings UI
- Refresh actions are available for each advanced setting
- App compiles cleanly and localized string keys exist for supported locales
- No unrelated behavior changes are included

zjs81:
Substantial repeater settings additions have landed: path hash mode (1/2/3-byte paths), GPS toggle, channel utilization stats, auto clock sync after login, and more. Could you check the current Repeater Settings screen and confirm whether the specific controls you needed are now present, or describe what's still missing?

PR #238, opened 2026-02-25
Add advanced repeater settings controls to repeater screen

Specter242:
## Summary
Adds an advanced settings section to the repeater settings screen and wires refresh actions for the new controls.

## Scope
- `lib/screens/repeater_settings_screen.dart`
  - Add advanced repeater settings controls and related refresh actions.
- `lib/l10n/app_en.arb`
  - Add new English string keys for advanced repeater settings labels/actions.
- `lib/l10n/app_localizations*.dart`
  - Add generated localization accessors/implementations for the new keys across locales.

## Included Controls
- Interference threshold
- AGC reset interval
- Flood max hops
- Multi ACKs

## Validation
- `flutter analyze lib/screens/repeater_settings_screen.dart lib/l10n/app_localizations.dart lib/l10n/app_localizations_zh.dart` (pass)

## Issue
Refs #237


lib/screens/repeater_settings_screen.dart
```diff
@@ -616,6 +723,22 @@ class _RepeaterSettingsScreenState extends State<RepeaterSettingsScreen> {
...
+      final agcResetSeconds =
+          int.tryParse(_agcResetIntervalController.text.trim());
+      if (agcResetSeconds != null) {
+        commands.add('set agc.reset.interval $agcResetSeconds');
+      }
+      final floodMax = int.tryParse(_floodMaxController.text.trim());
+      if (floodMax != null) {
+        commands.add('set flood.max $floodMax');
+      }
+      commands.add('set multi.acks ${_multiAcksEnabled ? 1 : 0}');
```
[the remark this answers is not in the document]
Specter242: Fixed in 7f38c9f. I added hydration/touched guards so `set multi.acks` is only sent when either (a) current value has been fetched from device, or (b) the user explicitly toggled the switch in this session. This prevents unintended reset to 0 when saving unrelated fields.

lib/screens/repeater_settings_screen.dart
```diff
@@ -44,6 +49,19 @@ class _RepeaterSettingsScreenState extends State<RepeaterSettingsScreen> {
...
       TextEditingController();
 
+  // Advanced settings
+  final TextEditingController _intThreshController = TextEditingController(
+    text: '14',
+  );
+  final TextEditingController _agcResetIntervalController =
+      TextEditingController();
+  final TextEditingController _floodMaxController = TextEditingController(
+    text: '4',
```
just-stuff-tm: Blocking: `_saveSettings()` always emits `set int.thresh` and `set flood.max` because the controllers are pre-seeded (`14`/`4`). This can overwrite existing repeater values even when user did not touch advanced settings. Please gate these like `multi.acks` (hydrate/touched flags) or initialize empty and only send when user changed or after successful fetch.

lib/l10n/app_en.arb
```diff
@@ -1110,6 +1120,11 @@
   },
   "repeater_refreshBasicSettings": "Refresh Basic Settings",
   "repeater_refreshRadioSettings": "Refresh Radio Settings",
+  "repeater_refreshAdvancedSettings": "Refresh Advanced Settings",
+  "repeater_refreshInterferenceThreshold": "Refresh interference threshold",
+  "repeater_refreshAgcResetInterval": "Refresh AGC reset interval",
+  "repeater_refreshFloodMaxHops": "Refresh flood max hops",
+  "repeater_refreshMultiAcks": "Refresh multi ACKs",
```
just-stuff-tm, line 1127: Please regenerate locale translations before merge. The generated locale files currently contain English fallback text for new repeater advanced settings keys.

Steps to fix:

1. Ensure Ollama is running and pull the expected model:
```bash
ollama pull translategemma:latest
```

2. Translate missing keys from `app_en.arb` into all locale ARBs:
```bash
python3 tools/translate.py --in lib/l10n/app_en.arb --l10n-dir lib/l10n --missing-only
```

3. Regenerate localization Dart files:
```bash
flutter gen-l10n
```

just-stuff-tm:
@codex review

just-stuff-tm (changes requested):
### Findings

* `_intThreshController` defaults to `14` and `_floodMaxController` defaults to `4`.
  `_saveSettings()` always sends them if they are parseable. Opening settings and saving unrelated fields can unintentionally reset device values.

---

* Every refresh clears and reuses the single `_fetchedSettings` map.
  If the user triggers multiple refresh actions quickly, one refresh can clear data while another is still collecting.

---

* Save-path validation for new numeric advanced fields is weaker than fetch-path validation.

  Fetch validation enforces a non-negative integer format, but save only performs `int.tryParse` and sends the value directly. Invalid or negative pasted values can still be transmitted.

---

* Regenerate Localization dart files after running translate.py

--- 

Please fix these so upstream can merge...   Thank You

zjs81:
We need to modify the repeater settings so it only sends modified fields.

Specter242:
I'll work on this in the next few days.  I'm on a job for the moment.

Specter242:
Addressed the requested follow-ups in 8954379.

Changes included:
- repeater settings now only send fields modified in the current session
- removed default advanced-value overwrites and gated saves behind actual dirty state
- isolated refresh fetch state per request to avoid cross-refresh races
- tightened numeric validation on save for the advanced fields and TX power
- updated locale ARBs, regenerated l10n files, and cleared the stale untranslated report

Verification:
- flutter gen-l10n
- flutter analyze lib/screens/repeater_settings_screen.dart lib/l10n/app_localizations.dart lib/l10n/app_localizations_zh.dart

Specter242:
The remaining failed check is the  workflow's test step, and it appears to be an existing upstream failure rather than something introduced by this branch.

Current PR failure:
-  /  on March 15, 2026 at 03:14 UTC
- 5 failing tests in 
- failures are  for  and the related  follow-on errors

Same failure is already present on :
- workflow run  on March 15, 2026 at 01:49 UTC
- same 5 failures in 

So the PR-specific changes are pushed and green on build targets, but the shared TCP test failure is still blocking the  check.

Specter242:
The remaining failed check is the `Flutter and Dart` workflow's test step, and it appears to be an existing upstream failure rather than something introduced by this branch.

Current PR failure:
- `analyze` / `Run tests` on March 15, 2026 at 03:14 UTC
- 5 failing tests in `test/screens/tcp_flow_test.dart`
- failures are `ProviderNotFoundException` for `AppSettingsService` and the related `Bad state: No element` follow-on errors

Same failure is already present on `main`:
- workflow run `23100927800` on March 15, 2026 at 01:49 UTC
- same 5 failures in `test/screens/tcp_flow_test.dart`

So the PR-specific changes are pushed and green on build targets, but the shared TCP test failure is still blocking the `Flutter and Dart` check.

446564:
sorry more merge conflicts

Specter242:
I'll get to it in a couple of weeks.  Thanks!

PR #239, opened 2026-02-26
Implements a debounced notification listener in MeshCoreConnector

wel97459:
Implement debounced notification listener updates in MeshCoreConnector.
This help with grouping UI rebuilds.


zjs81:
Wasn't this already a thing?

wel97459:
Not unless there is a system in place already that I didn't know about.

Issue #3, opened 2026-01-02
Enhancement Request: Show name of connected companion at top of screen


bplein:
When connected to the companion, it's not currently possible to see which companion you are connected to without going into Settings.

I propose that you put the name of the connected companion either up with the battery widget (similar to what MeshCore App does), or over on the right with the BT disconnect button.

MeshCore App example: 

<img width="422" height="318" alt="Image" src="https://github.com/user-attachments/assets/06d4ec5f-0cc9-4025-9b50-c2f96b360c23" />

446564:
I've been messing with placement and little ideas.

With the battery indicator seems like a good location but it has to get pretty small to fit the longest possible device name. And then it looks a little odd when the name is shorter.

<img width="1080" height="455" alt="Image" src="https://github.com/user-attachments/assets/7cca9867-c407-45fb-97db-56aefe0f716f" />

ericszimmermann:
It would be more aesthetically pleasing to have the brackets removed in https://github.com/zjs81/meshcore-open/blob/bdb1eb6b4210d351e83c41ac6054ccb29b0a78b0/lib/widgets/app_bar.dart#L43.
Otherwise this Issue is solved in current main branch.

ericszimmermann:
Completed, thank you.
Please close Issue.

446564:
complete

PR #240, opened 2026-02-26
Show name of connected companion

ericszimmermann:
A small change to close Issue #3.
I removed the brackets around the companion name, which I find aesthetically more pleasing.

Sincerely Eric


zjs81:
Thanks

Issue #255, opened 2026-03-03
Proposal for location based matching to Prefixes


ericszimmermann:
Hello I do not know flutter enough yet to program this,
but I wanted to make this proposal of my ideas.

I have two Issues:
- In SNR Indicator: Wrong name for repeaters! Prioritize Repeaters (Contacts should be direct contacts only) and on multiple prefix-matches on Repeater which is nearest.
- In Path Trace: Also on multiple matches to a prefix, take the location of the preceding and subsequent node. Take the point in between them and search from this point for the nearest prefix-match.

Here some moc-up code:
```c
location middle_location(contact prev_repeater, contact subsequent_repeater) {
  double avg_lon = (prev_repeater.location.lon + subsequent_repeater.location.lon)/2;
  double avg_lat = (prev_repeater.location.lat + subsequent_repeater.location.lat)/2;
  if( (prev_repeater.location.lon < -150 and subsequent_repeater.location.lon > 150) or ( (prev_repeater.location.lon > 150 and subsequent_repeater.location.lon < -150) ) ) {
    avg_lon = (prev_repeater.location.lon + subsequent_repeater.location.lon)/2;
    avg_lat = (prev_repeater.location.lat + subsequent_repeater.location.lat)/2;
    if( avg_lon <= 0 ) {
      avg_lon += 180.0;
    }
    else {
      avg_lon -= 180.0;
    }
  }
  return location(avg_lon, avg_lat)
}

contact find_nearest_contact(int unkown_repeater_prefixbyte, location poi) {
  double distance = DOUBLE_MAX;
  contact found_contact = None;
  int i = 0;
  
  for( i=0; i < len( contacts ); i++) {
    if contacts[i].matches_prefix(unkown_repeater_prefixbyte):
      double d = calc_distance( poi.lon, poi.lat, contacts[i] );
      if (d < distance ) {
        distance = d;
        found_contact = contact;
      }
    }
  }
}

contact find_nearest_repeater(int epeater_prefixbyte, contact prev_repeater, contact subsequent_repeater) {
  middle = middle_location(contact prev_repeater, contact subsequent_repeater);
  most_likely_repeater = find_nearest_contact( repeater_prefixbyte, middle );
  return most_likely_repeater
}
```

446564:
With multi-byte prefixes this will be far less of an issue, but still a nice feature if the user chooses to remain on single byte node hash size.

ericszimmermann:
[...]

PR #258, opened 2026-03-04
use location to match snr-indicator contacts

ericszimmermann:
Hi there,

as I wrote, I am not familiar with flutter programming, therefore I would like some feedback on this pull-request.
Then I used a bit of vibecoding, but only so I understood the made changes.

This Pull-request addresses the first part of my Ideas in Issue #255 to use location for the SNR-Indicator.
Instead of using the oldest matching prefix, I now use first contacts which have location data.
And on them if I have multiple matches (I had one), I use the nearest.
And if there are no contacts with location (maybe because its a direct connection to another companion) it should use the last-seen of the prefix matches.

Best regards
Eric

### DIFF
--- a/lib/utils/location_utils.dart
+++ b/lib/utils/location_utils.dart
@@ -7,7 +7,7 @@ Contact? selectBestRepeaterContactForPrefix(
   List<Contact> contacts,
   int pubkeyFirstByte, {
   LatLng? searchPoint,
-  bool preferFavorites = false,
+  bool? preferFavorites = false,
 }) {
   final candidates = contacts
       .where(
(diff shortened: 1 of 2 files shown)


ericszimmermann:
I tried to test it and it worked for this constellation with gps.
But I can not test all constellation, because when I readd the user I have no location data anymore.
Can someone tell me if the sorting and returning first of the list, when I do not have gps will give me the oldest or most recent entry? (Most recent should have a higher timestamp, therefore it should be correct.)

Screenshot_snr_test

ericszimmermann:
Tested it further at a break on my work location, where i have 2 more of these prefix situations.
One Repeater is blocked by a small mountain, so i know it should be the other one which is further away, because I can login with direct path. Therefore I implemented that the user has some control via Favorite flag.
Priorities are now Distance > Favorite > lastSeen.

ericszimmermann:
I am refactoring a litte bit to prepare for secand part of Issue #255.
Where I can use same function for both part.

ericszimmermann:
Is also included in #273 !

PR #262, opened 2026-03-06
fix: Chrome connect screen and handshake

benallfree:
I reworked the Chrome connection flow in a couple ways:

1. Changed the UI to be more intuitive since Chrome uses a native selector. There is no need for a `scan` button or a listing of BLE devices.

2. In the native Chrome BLE picker, we now filter by Nordic UART service ID because it is not possible to filter by `MeshCore-` prefix. tbh this seems like something we should probably do everywhere. I'm not sure why MC didn't choose their own service ID, it seems like that would clean up everything.

3. There is a bug in Flutter Blue that prevents the BLE handshake from working correctly on Chrome. That is fixed with a patch.

image

image

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -686,6 +686,11 @@ class MeshCoreConnector extends ChangeNotifier {
   }) async {
     if (_state == MeshCoreConnectionState.scanning) return;
 
+    if (kIsWeb) {
+      await _startScanWeb(timeout: timeout);
+      return;
+    }
+
     _scanResults.clear();
     _setState(MeshCoreConnectionState.scanning);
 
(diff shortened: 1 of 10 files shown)


zjs81:
The UI should be consistant on all platforms. Doing this adds additional complexitiy so I dont think its worth it atm.

benallfree:
Hmm, let me see if I can split out the handshake fix. Sorry I mixed them together.

benallfree:
Just to be clear - the BLE connection ux on Chrome can not be the same anyway. 

A native BLE selector is the *only* way to select a device. The app UI is redundant because all it does is list the one you already just chose and make you click 'connect' again. Also, the 'scan' button does not function on Chrome. All it does is pop the native BLE selector.

446564:
I agree that each platform will slowly diverge anyway and they are different systems. It would be clunky and misleading to have scan or a device list.

446564:
please resubmit just the handshake fix and have a look at the [contribution guide](https://github.com/zjs81/meshcore-open/blob/dev/CONTRIBUTING.md) if you haven't already

PR #265, opened 2026-03-06
enh: Chrome UI performance

benallfree:
PR stack:

- #262 

During initial loading, the UI is a little janky on Chrome. I made some adjustments to reduce the number of re-renders and to display Channels as they arrive rather than waiting until the end.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -152,6 +152,9 @@ class MeshCoreConnector extends ChangeNotifier {
   bool _isLoadingContacts = false;
   bool _isLoadingChannels = false;
   bool _hasLoadedChannels = false;
+  Timer? _contactLoadNotifyTimer;
+  bool _hasPendingContactLoadNotify = false;
+  final Set<String> _pendingContactMessageLoads = {};
   bool _batteryRequested = false;
   bool _awaitingSelfInfo = false;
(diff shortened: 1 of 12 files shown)


benallfree:
This one may have the same issue as #262 where it modifies the UI slightly for better performance.

446564:
this also touches a few too many things like the meshcore connector. each change would ideally be discussed prior and have a separate PR with one commit per the contribution guide.

benallfree:
FYI this is a [stacked PR](https://www.stacking.dev/) so changes might appear unrelated because they are present in a base PR.

446564:
> FYI this is a [stacked PR](https://www.stacking.dev/) so changes might appear unrelated because they are present in a base PR.

each PR should be able to apply cleanly to dev, so in a case like that it is best to get the base change in first, then submit the next change after it is in dev.

benallfree:
copy. doesn't work as well across forks, but when the base change goes in the diffs will drop out. in the mean time it's hard to read sorry

446564:
> copy. doesn't work as well across forks, but when the base change goes in the diffs will drop out. in the mean time it's hard to read sorry

ya it's all good, and squash and merge will work going forward. just try to split up changes and discuss before hand so you can get an idea of the direction to go.

we are on discord and irc if that's of use.

PR #277, opened 2026-03-11
Refactor storage classes to include companion's public key

wel97459:
This adds support to connect to multiple companions without having the data bleed over into each other.

Adds 5 bytes of the companion's public key to the prefix of the data stores.


wel97459:
@codex review

PR #285, opened 2026-03-11
Normalize ack packet parsing to firmware docs

robekl:
## Problem

The implementation treated `PACKET_MSG_SENT` timeout values and `PACKET_ACK` layout according to older assumptions:
- the send timeout was treated as milliseconds instead of documented seconds
- acknowledgements were assumed to always carry a 4-byte hash plus trip time

The firmware docs describe a different contract, including a shorter ACK variant. That mismatch can distort retry timing and make ACK handling incompatible with documented firmware behavior.

## Fix

Add dedicated parsers for:
- message-sent packets, converting the documented timeout units into milliseconds for internal use
- ACK packets, supporting both the documented short ACK form and the extended variant with trip time

Refactor connector handling to use those parsers instead of hard-coded offsets.

Add tests that verify:
- documented timeout conversion
- parsing of the documented short ACK variant
- parsing of the extended ACK variant

## Why this fixes it

Retry and delivery handling now normalize documented wire values into the app's internal model instead of assuming a single legacy frame shape. That preserves internal behavior while accepting the protocol formats described by firmware docs, which improves interoperability and timing correctness.

## Tradeoffs

For compatibility with existing internal ACK tracking, the short ACK variant is reduced to the first 4 bytes where the rest of the app still expects a 4-byte key. That keeps the change bounded, but it means the app is not yet using the full 6-byte ACK value end-to-end.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -3275,15 +3275,10 @@ class MeshCoreConnector extends ChangeNotifier {
   }
 
   void _handleMessageSent(Uint8List frame) {
-    // Frame format from C++:
-    // [0] = RESP_CODE_SENT
-    // [1] = is_flood (1 or 0)
-    // [2-5] = expected_ack_hash (uint32)
-    // [6-9] = estimated_timeout_ms (uint32)
-
-    if (frame.length >= 10) {
-      final ackHash = Uint8List.fromList(frame.sublist(2, 6));
-      final timeoutMs = readUint32LE(frame, 6);
+    final packet = parseMessageSentPacket(frame);
+    if (packet != null) {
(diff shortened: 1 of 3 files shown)


wel97459:
The time is in milliseconds, not in seconds, for the timeout.

`    if (recipient.out_path_len == OUT_PATH_UNKNOWN) {
      sendFloodScoped(recipient, pkt);
      est_timeout = calcFloodTimeoutMillisFor(t);
      return MSG_SEND_SENT_FLOOD;
    } else {
      sendDirect(pkt, recipient.out_path, recipient.out_path_len);
      est_timeout = calcDirectTimeoutMillisFor(t, recipient.out_path_len);
      return MSG_SEND_SENT_DIRECT;
    }`

PR #286, opened 2026-03-11
Interpret battery packets as documented percent values

robekl:
## Problem

The app interpreted battery packets as millivolts and derived percentage from voltage heuristics. The firmware docs describe that packet as a battery percentage value, optionally followed by storage counters. If the docs are authoritative, the UI was displaying incorrect battery information and mixing protocol parsing with device-specific voltage estimation.

## Fix

Parse the battery packet according to the documented percentage format and store the reported percentage directly. Keep voltage-dependent UI behavior behind an explicit "has voltage" check so the app still renders cleanly when only percentage is available.

Add tests that verify:
- documented battery packet parsing
- optional storage parsing behavior
- UI rendering of battery percentage without requiring a voltage value

## Why this fixes it

The app now displays the battery value the firmware actually reports instead of inventing a voltage interpretation from the same bytes. That makes the battery UI consistent with the documented protocol and removes chemistry-specific estimation from a packet that is already a percentage.

## Tradeoffs

This change gives up inferred voltage display for this packet format. If some older firmware really emitted millivolts on the same opcode, that legacy interpretation is no longer the primary path. The tradeoff is intentional because protocol correctness is more important than preserving an undocumented decoding heuristic.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -159,6 +159,7 @@ class MeshCoreConnector extends ChangeNotifier {
   bool? _clientRepeat;
   int? _firmwareVerCode;
   int? _batteryMillivolts;
+  int? _batteryPercentValue;
   double? _selfLatitude;
   double? _selfLongitude;
   final List<DirectRepeater> _directRepeaters = List.empty(growable: true);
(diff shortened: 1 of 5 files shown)


wel97459:
This is not valid for what the firmware actually sends.

` 
} else if (cmd_frame[0] == CMD_GET_BATT_AND_STORAGE) {
    uint8_t reply[11];
    int i = 0;
    reply[i++] = RESP_CODE_BATT_AND_STORAGE;
    uint16_t battery_millivolts = board.getBattMilliVolts();
    uint32_t used = _store->getStorageUsedKb();
    uint32_t total = _store->getStorageTotalKb();
    memcpy(&reply[i], &battery_millivolts, 2); i += 2;
    memcpy(&reply[i], &used, 4); i += 4;
    memcpy(&reply[i], &total, 4); i += 4;
    _serial->writeFrame(reply, i);
}
`

PR #287, opened 2026-03-11
Serialize startup commands by response

robekl:
## Problem

The firmware docs require sending one command at a time and waiting for the matching response before issuing the next command. The app's startup handshake sent several requests back-to-back during connection and refresh flows. That can race responses, violate documented sequencing expectations, and make startup behavior depend on firmware tolerance for concurrent outstanding commands.

## Fix

Introduce documented command-response matching for the startup sequence and serialize the initial handshake commands:
- device query
- app start
- custom vars
- battery/status
- auto-add config

Add a small helper that sends a command, waits for the expected response type, and times out cleanly before continuing.

Add tests that verify the documented response mapping and frame-to-command matching behavior.

## Why this fixes it

Startup now follows the sequencing model described by the firmware docs. Each request is allowed to complete before the next one is sent, which removes command overlap during the initial handshake and makes the app's connection flow conform to the documented command queue requirement.

## Tradeoffs

This makes startup slightly more serialized and can increase connection latency on slow links because commands are no longer pipelined. That is the correct tradeoff for the handshake path because reliability and protocol conformance matter more than shaving a small amount of startup time.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -1409,6 +1409,70 @@ class MeshCoreConnector extends ChangeNotifier {
     await sendFrame(buildGetBattAndStorageFrame());
   }
 
+  Future<bool> _sendFrameAndWaitForExpectedResponse(
+    Uint8List command, {
+    Duration timeout = const Duration(seconds: 5),
+  }) async {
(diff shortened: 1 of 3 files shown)


wel97459:
@codex review

wel97459:
The companion buffers the data in the order it is received.

PR #288, opened 2026-03-12
Integrate firmware protocol compatibility follow-ups

robekl:
## Summary

This integration branch rolls up the protocol-alignment work from the split firmware-doc fix branches and the follow-up compatibility fixes discovered during device testing.

Major areas covered:
- align app startup, reboot, channel, battery, ACK, BLE buffering, and startup sequencing behavior with the documented firmware protocol
- harden protocol builders, parsers, frame buffering, and validation around malformed or truncated data
- improve BLE/iOS runtime behavior, startup gating, login handling, diagnostics, and repeater flows discovered during on-device testing
- add broad regression coverage for protocol parsing, frame splitting, login outcomes, ACK handling, time-sync policy, retry policy, and UTF-8 truncation behavior
- migrate remaining packet/frame parsing paths onto `BufferReader` for consistent sequential decoding

## Included behavior changes

Protocol compatibility and startup:
- align `CMD_APP_START` with the documented wire format
- route reboot through the documented CLI path
- align channel frame encoding/parsing with documented secret handling, including truncated channel info responses
- interpret battery packets as documented percentage/status payloads
- normalize sent/ack packet parsing to documented frame shapes and timeout units
- serialize startup request/response handling where the firmware expects one-at-a-time command flow
- treat battery/custom-vars/auto-add startup requests as best-effort where firmware timing is loose
- skip automatic startup time sync on firmware `10+`, where the device rejects that write

BLE/frame parsing/runtime:
- buffer fragmented BLE notifications safely and stop emitting synthetic garbage packets
- correctly treat `RESP_CODE_CONTACTS_START` / `RESP_CODE_END_OF_CONTACTS` as 5-byte frames
- split trailing `RESP_CODE_OK` frames cleanly from adjacent fixed-length packets
- fix contact parsing so full 148-byte frames are consumed and direct contacts stay `pathLength == 0`
- remove startup auto-writeback of auto-add/other params
- improve command/error attribution and stale pending-command cleanup for overlapping async responses
- disable unstable iOS BLE restoration after it regressed real iPhone scan/connect behavior

Repeater/login/remote command handling:
- accept compact one-byte login outcome frames
- sanitize repeater timeout math when firmware reports invalid radio params
- make login wait for explicit login success/fail outcomes instead of treating generic `ERR` as terminal failure
- improve repeater/room login diagnostics and preserve useful local validation errors
- stop CLI/admin traffic from polluting chat ACK matching
- handle binary response packets centrally instead of logging them as unknown frames

Parser/builder hardening:
- validate malformed keys, timestamps, radio params, coordinates, payload sizes, and frame size limits across protocol builders
- harden `BufferReader` string readers against truncation and boundary issues
- reject malformed ACK/battery/sent packet sizes instead of accepting ambiguous data
- prevent UTF-8 truncation from splitting multibyte characters in outbound names/labels
- force flood-routed retry attempts onto the documented attempt value

`BufferReader` migration requested in review:
- shared protocol parsers now use `BufferReader` for battery, sent, ACK, login outcome, binary response, and contact text parsing
- model frame constructors now use `BufferReader` for:
  - `Contact.fromFrame`
  - `Channel.fromFrame`
  - `Message.fromFrame`
  - `ChannelMessage.fromFrame`
- screen-level packet decoders now use `BufferReader` for:
  - repeater binary status parsing
  - BLE raw-log packet, advert, and control summary decoding
- added `BufferReader` helpers for bounded peeking, remaining-length checks, and fixed-width padded C-string parsing

## Testing

Validated on this branch with:
- `flutter analyze`
- `flutter test`
- `flutter build ios --debug --no-codesign`
- on-device iPhone BLE validation against MeshCore firmware v9 and v10 for:
  - startup handshake
  - contacts and channel sync
  - repeater login
  - status/binary requests
  - CLI commands
  - trace-path behavior

Additional regression coverage added on this branch includes:
- `BufferReader` fixed-width/padded string parsing
- message and channel-message frame parsing via `BufferReader`
- command/response matching
- frame-buffer edge cases
- ACK and generic-ACK queue handling
- login outcome parsing
- battery/status parsing
- protocol math and codec boundaries


robekl:
Logfile analysis of upstream/dev on iPhone when reverting after running the changes on this branch:

1. `RESP_CODE_OK` is not handled at all in the connector, so every normal `OK` reply is being mislabeled as an unknown frame. The log repeatedly shows `RX frame: code=0 len=1` followed by `Unknown frame code: 0`. `respCodeOk` is defined in [meshcore_protocol.dart](lib/connector/meshcore_protocol.dart#L186), but there is no matching `case` in the receive switch in [meshcore_connector.dart](lib/connector/meshcore_connector.dart#L1718) and the default path logs it as unknown at [meshcore_connector.dart](lib/connector/meshcore_connector.dart#L1805).

2. ACK matching is still being applied to traffic that is not a tracked outbound chat send. The repeated sequence `RX frame: code=6 len=10`, `Hash-based match failed ...`, `No pending message found for ACK hash ...` means `RESP_CODE_SENT` frames from login/status/binary/CLI operations are still flowing into chat ACK correlation. The matcher behavior is in [message_retry_service.dart](lib/services/message_retry_service.dart#L237), and the connector is still dispatching `respCodeSent` without enough filtering in [meshcore_connector.dart](lib/connector/meshcore_connector.dart#L1766).

3. `PUSH_CODE_BINARY_RESPONSE` is still unhandled at the connector level on this branch, which is why the log shows `RX frame: code=140 len=18` followed by `Unknown frame code: 140`. The code is defined in [meshcore_protocol.dart](lib/connector/meshcore_protocol.dart#L218), and some screens parse it directly in [telemetry_screen.dart](lib/screens/telemetry_screen.dart#L56) and [neighbors_screen.dart](lib/screens/neighbors_screen.dart#L57), but the central receive switch still falls through to unknown-frame logging in [meshcore_connector.dart](lib/connector/meshcore_connector.dart#L1805).

4. Startup sequencing still allows channel sync to overlap the initial handshake/contact load. The log shows channel sync starting immediately after device info, then later `Already syncing channels, ignoring request`. The relevant gating is in [meshcore_connector.dart](lib/connector/meshcore_connector.dart#L1546) and [meshcore_connector.dart](lib/connector/meshcore_connector.dart#L1884). It is not breaking outright here, but it is still noisy and race-prone.

Residual notes:
- `MTU request failed ... android-only` is expected on iOS.
- The CoreBluetooth restore warning is plugin/config noise unless you expect restoration.
- `RX frame: code=13 len=82` suggests firmware/device-info drift, but nothing in this log proves that extra byte is breaking parsing.

wel97459:
Use the BufferReader Class to parse packets.

446564:
please review and resolve conlficts

446564:
honestly this should be split up, PR should do one thing and in a single commit ideally.

broad or major changes should first be discussed, not just a huge PR sent over. You can refer to the https://github.com/zjs81/meshcore-open?tab=contributing-ov-file#readme for details.

PR #291, opened 2026-03-13
Dev discovery

wel97459:
Discovered contacts use the normal contacts class.
The Contacts class has a new bool, isActive, that indicates whether the contact is in the contacts list.

Discovered contacts are combined in the map view, path tracing, and SNR indicator.

Discovered contacts are accessible between companions to make it easy to add contacts to other companions.

Also fixes my mistake in the data stores.


wel97459:
@codex review

PR #293, opened 2026-03-14
feat: add set-as-my-location from map long-press, connector and UI 

zjs81:
improvements

Add "Set as my location" option to the map long-press bottom sheet, allowing users to set their device position directly from the map. Includes connector, chat, contacts, and message retry service improvements.


zjs81:
@codex review

PR #298, opened 2026-03-15
import export discovered

ericszimmermann:
Allows to import and export discovered Contacts

I hope it is easier to manage if i split up my changes.


ericszimmermann:
I let my agent change the detection from overlapping prefixes change from an O(n^2) to O(n) approach and also added this and the guessed location into app-settings to allow disabling it before going to map.
Otherwise the app would have been to slow to handle my 1200+ discovered contacts.

With the change of the prefixes the map could handle the 1200+ contacts.

PS: I do not know why dart says LatLang is an unused import, since it clealy does not build without."

PR #316, opened 2026-03-26
Make unread badge max out at 9999+ instead of 99+

n-kam:
Tiny change. Public chat on large networks can often produce couple of hundreds messages per day. So '99+' is not so informative.


zjs81:
Thank you!

Issue #279, opened 2026-03-11
Add 'chats' tab


446564:
An nice idea would be to have the first tab be chats, most people are using meshcore to chat with individuals or channels. So if we had: Chats, Channels and Map. When on the Chats tab there could be a `+` button that would bring up contacts, or there could be an additional tab for all contacts as there is now.

just-stuff-tm:
i like this idea and the chat with most recent message on top

ericszimmermann:
Maybe it's more of a filter: only chats with more than one message?

ericszimmermann:
I made an implementation of my proposal.
Please try and give feedback.
Best Regards, Eric

PR #320, opened 2026-03-28
Chat Filter

ericszimmermann:
Hi there,

I added a "Active Chat" filter, which Displays all Chats which have at least one Message. (according to my proposal in Issue #279)
To delete from List you will have to go into the chat and "Clear Chat" in the Filter menu.

Best Regards
Eric Zimmermann

PS: I also gave a little more space to the search bar.

### DIFF
--- a/lib/l10n/app_bg.arb
+++ b/lib/l10n/app_bg.arb
@@ -1353,6 +1353,7 @@
   "listFilter_users": "Потребители",
   "listFilter_repeaters": "Повторители",
   "listFilter_roomServers": "Сървъри на стая",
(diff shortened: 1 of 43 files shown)


446564:
I think #279 has some better solutions, instead of a filter which we already have with sorting by latest messages and combined with filtering unread only.

ericszimmermann:
Do you want me to close it?

446564:
just waiting on feedback from the rest of the team, hang on to it for now

446564:
I think we are going to go with the tab option, exact layout etc tbd

PR #326, opened 2026-03-30
Better french translations

spfmoby:
As usual some improvements in the french translations.

I noticed that the "3 dots -> Discovered Contacts" is not translated, even if the translation is properly done in the french file. Hard corded string?


zjs81:
Thank you, I will create an issue for that.

PR #321, opened 2026-03-28
Add additional device name prefixes to MeshCoreUuids

just-stuff-tm:
This pull request makes a minor update to the `MeshCoreUuids` class by adding two new device prefixes to the list of recognized device names.

- Added `"Seeed"` and `"Lilygo"` to the list of device name prefixes in `MeshCoreUuids` to support additional hardware.

PR #337, opened 2026-04-05
Add LowMesh prefix and explain how to add more

interfect:
This adds the Bluetooth name prefix used by the LowMesh Pocket-Ranger (and possibly other LowMesh brand devices) when running LowMesh's flavor of MeshCore 1.14 as obtained from [the LowMesh flasher](https://lowmesh.com/flasher).

It also updates various places in the documentation that referenced only one or a couple prefixes, and documents where the real list of prefixes lives and how to add more.

I tested this on Android, and Alpha 7 couldn't see the device whereas this build (eventually) could. (With the Windows build I could see the device whether the name prefix was in the list or not, but could never connect to it.)

This hopefully won't conflict with #321, since the added lines are in slightly different places.


zjs81:
Thank you.

Issue #336, opened 2026-04-04
Contact Groups sometimes removes nodes GPS coordinates


Snayler:
I created a few contact groups, one of them named neighbors and all the contacts I added to it lost their GPS position.

I tried to replicate this problem and couldn't figure out exactly what is causing it, but I managed to make 2 groups that made the contacts loose their gps coordinates.

ericszimmermann:
I noticed something similar, but I do not think it has to do with the groups. I think it has to do more with reloading the contactlist from the companion and consolidating with the discovered list.
Sincerely Eric

Snayler:
Interesting. I did reload the contacts inside the group and only those contacts lost the GPS info. I'll have to test it out some more

ericszimmermann:
I asked the AI, thats what I got as response:

**How Groups Work**
Groups are a pure UI filter — ContactGroup only stores memberKeys (a list of publicKeyHex strings). When a group is selected, _filterAndSortContacts in contacts_screen.dart:877 simply filters connector.contacts using:
```c
filtered = filtered
    .where((contact) => memberKeys.contains(contact.publicKeyHex))
    .toList();
```
The Contact objects returned are the exact same objects from connector.contacts — there's no copying, transforming, or stripping of data. So groups themselves cannot cause GPS/path loss.

**The Real Bug: GPS Wiped on Contact Refresh**
There's a data loss issue in meshcore_connector.dart:3920 that affects all contacts (and would be noticed in groups):
```c
if (latRaw != 0 || lonRaw != 0) {
    lat = latRaw / 1e6;
    lon = lonRaw / 1e6;
}
// Otherwise lat/lon remain null
```
When a contact update arrives via respCodeContact or pushCodeNewAdvert, the frame parsed by Contact.fromFrame sets latitude/longitude to null if the device sends lat=0, lon=0:

Then in _handleContact, the stored contact is replaced with:
```c
_contacts[existingIndex] = contact.copyWith(
  lastMessageAt: mergedLastMessageAt,
  pathOverride: existing.pathOverride,     // ✓ preserved
  pathOverrideBytes: existing.pathOverrideBytes, // ✓ preserved
  // ❌ latitude/longitude NOT preserved – if new contact has null, existing GPS is wiped
);
```
Since copyWith uses contact.latitude ?? this.latitude, and contact.latitude (from the new frame) is null, the result is null regardless of whether existing had GPS data.

Fix: preserve existing GPS when the device doesn't send any:
```c
[...]

PR #339, opened 2026-04-05
Preserve Coordinates with contact.copyWith() function

ericszimmermann:
Previously, the merge only preserved path override fields and could overwrite existing GPS with null when the incoming frame had 0,0 coordinates. Now it also preserves prior coordinates when the incoming update omits location.
Should address Issue #336.


zjs81:
Thank you

Issue #341, opened 2026-04-06
Add graceful degradation to GIF message encoding


interfect:
Right now, when I send a GIF, users of clients that don't understand GIF messages do not have a way to:
1. Know that what they are looking at is a GIF message
2. Access the content of the message (i.e. somehow manually look up the GIF).

It would be good if these were a solution that allowed people to do one or both of these.

My recommended approach would be to change the GIF encoding to use a URL to something. Since it looks like the GIFs already come from Giphy, we could use the full Giphy URLs instead of just the IDs being used currently.

The downside is that this would take more bytes per image. But it would still, as far as I can tell, fit comfortably under the message limit, and it would mean that basically every client would be able to get *something* useful out of the message, so it would probably *reduce* concerns like those in #109 about "wasting" airtime or "spamming" the mesh when these extended-feature messages are sent.

interfect:
It looks like, for a GIF ID like `MahDrOWLffiMKpiVV0`, the URL used for display right now is `https://media.giphy.com/media/MahDrOWLffiMKpiVV0/giphy.gif` and the shortest working URL I can find is `https://giphy.com/gifs/MahDrOWLffiMKpiVV0`. That immediately redirects to a longer URL that gives you the main Giphy page *about* the GIF, with the GIF on it.

I'm going to try teaching the client to start *understanding* some or all of these URLs the same way it understands the `g:` messages, and PR that. Then if one of these is adopted (here or elsewhere), it will render correctly.

zjs81:
Addressed in Alpha8 via PR #342 ('Support receiving more formats of GIF message'), which added support for multiple Giphy URL formats and centralised GIF parsing for graceful handling of varied encodings. Closing as resolved.

PR #342, opened 2026-04-06
Support receiving more formats of GIF message

interfect:
This does about half of #341. In addition to `g:GIFID` messages, this allows parsing messages consisting entirely of a single Giphy URL as GIF messages, by extracting the GIF ID from the URL.

To do this, I unified the three copies of GIF message parsing into one, in a new `GifHelper` class modeled on the `ReactionHelper`.

I also gave both the `GifHelper` and `ReactionHelper` the responsibility of *encoding* as well as parsing their message types, since it seemed to make the most sense to have the encoding and decoding sides together where they can both be worked on at once.

### DIFF
--- a/lib/services/translation_service.dart
+++ b/lib/services/translation_service.dart
@@ -6,6 +6,7 @@ import 'package:llamadart/llamadart.dart';
 
 import '../models/app_settings.dart';
 import '../models/translation_support.dart';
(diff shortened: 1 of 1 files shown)


zjs81:
@codex review

zjs81:
Very Nice! I will do some testing and get it merged in a few hours. Thank you.

PR #343, opened 2026-04-06
Send GIFs as more verbose but more compatible clickable links

### LINKED ISSUE #341: Add graceful degradation to GIF message encoding

interfect:
This builds on top of #342 and adds a commit to change the *sent* format of GIF messages from `g:GIFID` to `https://giphy.com/gifs/GIFID`.

This would close #341.

I'm not sure how people like this, because it adds 22 bytes to every GIF message versus the current format, and (unlike #342) the additional benefits would be felt only by users of *other* clients, who would be able to click the links and see the GIFs.

This could be shaved down to 21 bytes by using `http://` for the protocol, and 13 bytes by dropping the protocol altogether (just sending `giphy.com/gifs/GIFID`), which would still be human-readable but would likely not be clickable in most clients.

I'm going to try running this and see if my local mesh community is less baffled by my constant GIFing.

### DIFF
--- a/lib/helpers/gif_helper.dart
+++ b/lib/helpers/gif_helper.dart
@@ -33,6 +33,6 @@ class GifHelper {
   
   /// Encode a GIF in a format that parseGif() can parse.
   static String encodeGif(String gifId) {
-    return 'g:$gifId';
+    return 'https://giphy.com/gifs/$gifId';
   }
 }


zjs81:
I appreciate the PR but I dont think this is needed at the moment. If someone is confused they could just ask what it is.

PR #346, opened 2026-04-07
Location Sharing and marker deduplikation.

ericszimmermann:
Created a Popup for the input field, where you can select to:
- Add Gif (as bevore)
- Insert Emoji from internal Emoji Picker
- Share Location as marker.

The Share Location is only enabled if GPS is enabled on companion.
It has two modes:
1) single send of current location: to share your position. You can edit the Label for this.
2) automated sending of location in certain situation, for limited time (max 24h):
  It only works for Contacts or PRIVATE channels (and would be best to limit per scope as soon as thats implemented):
  For Private channels only sends when location has changed, and does no retry.
  For Contacts it will try to use path and if this is not working use a single floot message in the next interval.
  If you loose Bluetooth connection, sharing will be aborted. The label of the marker will always be the Companion name.

For this to work markers get now updated instead of only added (sender, label and flags have to be the same)

Since working on the input field, I also modified the length limitter to allow full size on smaz compression.

Sincerely Eric


ericszimmermann:
Testbuild with all four Pullrequest from me can be found here:
[https://github.com/ericszimmermann/meshcore-open/actions/runs/24091289280/artifacts/6310045189](https://github.com/ericszimmermann/meshcore-open/actions/runs/24091289280/artifacts/6310045189)

446564:
it might be better to break this into smaller changes. one commit per PR, one coherent change.

so one for smaz aware utf length limit, one for share location, one for emoji picker etc etc. get the one any others rely on merged first then build on dev again for the next. it will make getting things merged in much quicker.

ericszimmermann:
I will try it, but it gets a little complicated because they depend on each other.
This is why this was changed back to draft.

446564:
> I will try it, but it gets a little complicated because they depend on each other. This is why this was changed back to draft.

ok maybe let's get the counter PR in. then maybe you can split the smaz aware part out and get that in, then the rest as one add new input to messages PR

446564:
given the conflicts and splitting out the functionality with the smaz PRs, let's close this and resubmit for the other changes. those should be discussed in an issue before hand.

PR #360, opened 2026-04-08
fix foreground service and add noitification nav

### LINKED ISSUE #58: [Feature Request Android]: Spawn foreground service
The app disconnects from the node if closed for a while because there's no foreground service to keep the system from killing the application.

446564:
wraps MaterialApp in WithForegroundService to keep alive when swiped away

persists last connected device and clears on manual disconnect to allow
reconnect after kill

added lifecycle tracking to iOS and keep android notification alive with
heartbeat

add notification navigation

closes #58 until we decide to change the whole architecture.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -40,6 +40,7 @@ import '../storage/contact_settings_store.dart';
 import '../storage/contact_store.dart';
 import '../storage/message_store.dart';
 import '../storage/unread_store.dart';
+import '../storage/last_device_store.dart';
 import '../utils/app_logger.dart';
 import '../utils/battery_utils.dart';
 import '../utils/platform_info.dart';
@@ -281,6 +282,7 @@ class MeshCoreConnector extends ChangeNotifier {
   final ContactDiscoveryStore _discoveryContactStore = ContactDiscoveryStore();
   final ChannelStore _channelStore = ChannelStore();
   final UnreadStore _unreadStore = UnreadStore();
+  final LastDeviceStore _lastDeviceStore = LastDeviceStore();
   List<Channel> _cachedChannels = [];
   final Map<int, bool> _channelSmazEnabled = {};
   bool _lastSentWasCliCommand =
@@ -768,6 +770,10 @@ class MeshCoreConnector extends ChangeNotifier {
     _appDebugLogService = appDebugLogService;
     _backgroundService = backgroundService;
     _timeoutPredictionService = timeoutPredictionService;
+
+    // When the app resumes from background, check if we need to reconnect.
+    _backgroundService?.onResume = _onAppResumed;
+
     _usbManager.setDebugLogService(_appDebugLogService);
     _tcpConnector.setDebugLogService(_appDebugLogService);
 
@@ -1879,6 +1885,7 @@ class MeshCoreConnector extends ChangeNotifier {
       );
 
       _setState(MeshCoreConnectionState.connected);
+      _lastDeviceStore.persistLastDevice(_deviceId!, _deviceDisplayName!);
       if (_shouldGateInitialChannelSync) {
         _hasReceivedDeviceInfo = false;
         _pendingInitialChannelSync = true;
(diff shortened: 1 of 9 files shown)


446564:
oh, we have tests now? and they test behavior not code lol

stphnrdmr:
> oh, we have tests now? and they test behavior not code lol

Somebody added these a couple of weeks ago 😆 
https://github.com/zjs81/meshcore-open/commit/ddee76ced26f41028f1b2dbc299ce07d09a20909

lib/services/background_service.dart
```diff
@@ -1,65 +1,159 @@
...
 
   @override
-  void onRepeatEvent(DateTime timestamp) {}
+  void onRepeatEvent(DateTime timestamp) {
+    // Periodically update the notification so the system considers the
+    // service active.  This also acts as a heartbeat.
+    FlutterForegroundTask.updateService(
+      notificationTitle: 'MeshCore running',
+      notificationText:
+          'Connected · ${timestamp.hour.toString().padLeft(2, '0')}:${timestamp.minute.toString().padLeft(2, '0')}',
```
stphnrdmr: The timestamp seems to be UTC, we should convert it to local time.
```suggestion
          'Connected · ${timestamp.toLocal().hour.toString().padLeft(2, '0')}:${timestamp.toLocal().minute.toString().padLeft(2, '0')}',

```

lib/services/background_service.dart
```diff
@@ -1,65 +1,159 @@
...
+  /// Update the persistent notification text (Android only).
+  Future<void> updateNotification({
+    required String title,
+    required String text,
+  }) async {
+    if (!PlatformInfo.isAndroid || !_serviceRunning) return;
+    FlutterForegroundTask.updateService(
+      notificationTitle: title,
+      notificationText: text,
+    );
```
stphnrdmr: While testing i noticed that the app doesn't seem to ask for notification permissions when installed. toggling notifications off and on immediately prompts for the permission. That prevents this from working properly.

lib/connector/meshcore_connector.dart
```diff
@@ -2225,6 +2231,78 @@ class MeshCoreConnector extends ChangeNotifier {
...
+      }
+    }
+  }
+
+  void _clearPersistedDevice() {
+    final prefs = PrefsManager.instance;
+    prefs.remove(_prefKeyLastDeviceId);
+    prefs.remove(_prefKeyLastDeviceName);
+  }
+
```
stphnrdmr: Do you want to extract this into a "store" similar to [lib/storage/channel_settings_store.dart](https://github.com/zjs81/meshcore-open/blob/main/lib/storage/channel_settings_store.dart)?

stphnrdmr (changes requested):
Looks good overall, just some small remarks. 👍

lib/services/background_service.dart
446564: ya that's next on my list today

446564:
still need to address tests

lib/connector/meshcore_connector.dart
```diff
@@ -4910,6 +4969,17 @@ class MeshCoreConnector extends ChangeNotifier {
         );
   }
 
+  /// Public accessor to find a channel by its index.
+  Channel? findChannelByIndex(int index) => _findChannelByIndex(index);
```
stphnrdmr, line 4973: Why not change the name of the original method?

lib/services/background_service.dart
```diff
@@ -1,65 +1,159 @@
...
+  }
```
stphnrdmr: Where is this method used? I can't seem to find a caller for it.

lib/services/notification_service.dart
```diff
@@ -8,6 +9,19 @@ import '../helpers/reaction_helper.dart';
...
+  /// One of: 'message', 'channel', 'advert', 'batch'.
+  final String type;
+
+  /// For messages: the contact public key hex.
+  /// For channels: the channel index as a string.
+  /// For adverts: the contact public key hex.
+  final String? id;
+
+  const NotificationTapEvent({required this.type, this.id});
+}
```
stphnrdmr, line 25: What do you think about using an enum here?

```suggestion
enum NotificationTapEventType { message, channel, advert, batch }

/// Payload emitted when the user taps a notification.
class NotificationTapEvent {
  final NotificationTapEventType type;

  /// For messages: the contact public key hex.
  /// For channels: the channel index as a string.
  /// For adverts: the contact public key hex.
  final String? id;

  const NotificationTapEvent({required this.type, this.id});
}
```

lib/services/background_service.dart
446564: That was an earlier thing I missed removing. I'll take it out.

lib/connector/meshcore_connector.dart
446564: just convention really, it's used in the same file privately but also exported for use outside the file

PR #361, opened 2026-04-08
remove unused macos path_provider_foundation

446564:
added in #299 but appears not needed, flutter removes when building


446564:
needs testing as I don't have a mac build environment

PR #365, opened 2026-04-10
enh: make repeater admin guest aware

### LINKED ISSUE #8: admin assumption
Heya - great work so far! 

As requested, raising this ticket. 

logging into repeaters doesn't know if I am guest or admin - and baseline assumes admin.

446564:
adds checks for the admin bit in the login response and a new pattern to exclude admin feature widgets etc

closes #8


446564:
ok pushed again with changed title and excluded battery chemistry info

just-stuff-tm (changes requested):
Image

Image


All looks good except there should be guest helpers and titles on the login screen and the repeater screen...

446564:
for the login it could just change to something like 'Enter the repeater password for guest or admin access' as we don't know if they are an admin until after login.

I'll reword a few more things on the hub screen, looks like some of those tiles should change too. like stats is not also neighbors etc. maybe that should be another PR though.

446564:
ok ready for final review

just-stuff-tm:
Ok I'll build and approve

untranslated.json
```diff
@@ -1,69 +1,120 @@
...
+    "repeater_guestTools"
   ],
 
   "zh": [
-    "chat_sendMessage"
+    "chat_sendMessage",
+    "repeater_guest",
+    "room_guest",
+    "repeater_guestTools"
   ]
```
just-stuff-tm, line 119: untranslated strings need to be translated before merge into main.... Good to merge into dev

just-stuff-tm (approved):
Good to merge into dev

zjs81 (approved):

Issue #349, opened 2026-04-08
[Feature Request] Auto time sync when managing repeater


fokcuk:
Adding auto time sync command to execute when connecting to a repeater as administrator 
If possible having an auto schedule in a client would also be nice, so that it auto connects and syncs the clock on one or multiple configurable repeaters

just-stuff-tm:
I made a feature branch for this will be working on implementing  [auto-time-sync-349](https://github.com/just-stuff-tm/meshcore-open/tree/feature/auto-time-sync-349)

just-stuff-tm:
Does anybody feel like this should have a toggle

446564:
this should be nice with an auto toggle, then you will sync each time you login.

zjs81:
Implemented in Alpha8 via PR #369 ('add auto clock synchronization setting after repeater login') — the repeater settings screen now has an auto clock sync toggle that runs automatically after each repeater login. Closing as resolved.

PR #369, opened 2026-04-10
add auto clock synchronization setting after repeater login

just-stuff-tm:
Introduced a new setting for automatic clock synchronization after a successful repeater login. Added localization support for the new feature in multiple languages (Bulgarian, German, English, Spanish, French, Hungarian, Italian, Japanese, Korean, Dutch, Polish, Portuguese, Russian, Slovak, Slovenian, Swedish, Ukrainian, Chinese). Implemented storage service methods to manage the new setting. Updated the repeater settings screen to include a toggle for the new feature. Enhanced the repeater login dialog to trigger clock synchronization automatically if the setting is enabled. [Feature Request #349](https://github.com/zjs81/meshcore-open/issues/349)

https://github.com/user-attachments/assets/b575dc54-a8f8-4a13-9f65-385c8d397013

### DIFF
--- a/lib/l10n/app_bg.arb
+++ b/lib/l10n/app_bg.arb
@@ -2061,5 +2061,13 @@
   "scanner_linuxPairingHidePin": "Скриване на PIN кода",
   "scanner_linuxPairingShowPin": "Покажи PIN",
   "repeater_cliQuickClockSync": "Синхронизация на часовника",
-  "repeater_cliQuickDiscovery": "Открий Съседи"
+  "repeater_cliQuickDiscovery": "Открий Съседи",
+  "@repeater_clockSyncAfterLogin": {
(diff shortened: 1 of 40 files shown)


just-stuff-tm:
@446564 @wel97459 @zjs81 @interfect 
@codex review

The scheduler can be implemented in a separate PR. Proper upstream error handling for clock sync should also be addressed separately.

The current error (“cannot turn clock back”) occurs because the clock is not actually ahead, it's in UTC, while the sync request is coming from UTC-4 or similar. There is no timezone conversion happening, which needs to be fixed in another PR as well.  

The OG app throws the same error and doesn't convert host time to UTC before sending to repeater

446564 (approved):

Issue #352, opened 2026-04-08
import/export discovered (reimplementation)


ericszimmermann:
I am considering to do a fourth implementation of the import/export functionality for the discovered contacts from #298.
Reimplementation happend due to big changes in main branch and/or convoluted codex change requests.
This time it should be more in Line with #347 (contributing.md) and will result to about 2 Pull-Requests.

This feature came with my desire to improve the drawing of paths of the channel messages in #307, but I still find it important.
Then for every implementation (most didn't make it to Github), I wanted to test with the same discovery list for better comparison.

The planned Implementation:
1) Prefix matching feature in map:
  The Implementation of this feature (mark matching prefixes red in map) is not feasable anymore.
  The Implementation is very calculative intense and is done on every move of the map. This makes it slow, and with about 1200 Repeaters can cause the App to freeze. In my previous Pull-Request I tried to migitate this through changing the calculation from n^2 to a precalculated table (n calculations).
  Now I propose to remove it completly, since with about 1500 Repeaters you have with 1byte Prefixes about 6 matches to a Prefix. This just colors all Repeatermarkers red.
 2) Import Export of Json:
    - Entries to Menu of Discovery screen (with Language files)
    - decoding of JSON with Emoji's (since I am not that familiar with special cases of UTF8/16, I have to use AI with that)

Please let me know your thoughts.

ericszimmermann:
This feature is now in its fifth implementation / rebase onto dev branch.
I use it on all my testbuilds and use it extensivly there, to transfer the discovered contacts, so that I can test the channel_message_path_screen or map performance with lots of repeaters (now around 3'000), without having to wait a week to be able to test.
None the less, I still got no responce. Therefore I ask, why the hold up?
It gets a little tedious to always keep it updated to dev branch, so what can I do to get it merged?

Sincerely Eric

PR #370, opened 2026-04-10
Remove colorisation of repeaters with same prefix as red

ericszimmermann:
This PR removes the feature to mark Repeaters with matching Prefixes as Red.

As also in Issue #352 explained, this is of no usage anymore:
This is due to the fact, that with the Discovery list (unlimited appstorage) you can get so many repeaters,
that you have multiple matches for every Repeater prefix, resulting just in marking every Repeater red without additional information gain.
For 1500 Repeaters and a one byte Prefix you get about 5 matches for every prefix.

Additionally the calculation is done in a repeater ^ 2 search method, and this not only once on map load, but on every map interaction (move/zoom).
(In my previous PR #298 I implemented a repeater ^ 1 approach to keep the functionality, but as written above I would now advertise to remove this.)

PS: I hope it is right to base it on dev-branch?

### DIFF
--- a/lib/l10n/app_bg.arb
+++ b/lib/l10n/app_bg.arb
@@ -1942,7 +1942,6 @@
   "path_routeWeight": "{weight}/{max}",
   "settings_multiAck": "Мулти-потвърди: {value}",
   "settings_telemetryModeUpdated": "Режим на телеметрията е обновен",
-  "map_showOverlaps": "Покриване на ключа на повтаряча",
   "map_runTraceWithReturnPath": "Върни се по същия път.",
   "@radioStats_noiseFloor": {
     "placeholders": {
--- a/lib/l10n/app_de.arb
+++ b/lib/l10n/app_de.arb
@@ -1970,7 +1970,6 @@
   "path_routeWeight": "{weight}/{max}",
   "settings_telemetryModeUpdated": "Telemetriemodus aktualisiert",
   "settings_multiAck": "Mehrfach-Bestätigungen: {value}",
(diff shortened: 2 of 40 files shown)


446564:
looks ok, testing now

446564 (approved):
no issues here

446564:
Looks like we aren't going to be getting rid of this, but perhaps changing the way we do it.

from @zjs81:
>I disagree. Realistically we should just do something like this 
>
>Prefixes = a1,a2,a1
>If (getprefix).count() > 1{
>Show red();
>}
>
>Then on click we get the repeaters lazily
>That way if that entry is in the list more than once we mark it red we can cache that
>So this pr could just fix it
>This is still a useful signal because duplicate prefixes can impact dm messages
>So we can just do this more efficiently instead of removing it
>We could even build a frozen collection if dart supports that

ericszimmermann:
Where is this quote from, because I seem to lack some of the context. He has a path from something? Or a selection of prefixes to mark red?

My point is, that if you have enough contacts, every possible prefix will have at least two matches, therefore just making the whole map red. With this implementation you can not even pair them up. If you have 4 repeaters, two with b7 and two with a3, all four will just show red. You can not distinguish wish is the alternative to the other, you would have to filter for the prefix, to get this information (another function, much more usefull for this).

Can there be given a usecase of when and how this function can be used?

Sincerely Eric

PS: I have of now 6 Repeaters on the Companion, which I interact with, the rest  of the 350 fields are auto filled with contacts. And about 1097 Repeaters in the discovery list with further 233 old (autodeleted) contacts. This amount to about 4 Repeaters to every possible prefix.

PR #373, opened 2026-04-10
import export function for discovered contacts

ericszimmermann:
Made the changes for importing and exporting Discovered contacts as JSON files.

This is according to Issue #352 (and based on now obsolete #298).

I also cleaned up the code a bit, and this is the isolated import/export functionality.

There was just on thing. there was a contactfield "wasPulled" which seemed to should have the same functionality as the existing "isActive" flag, but was not used anywhere, therefore I removed it. (I hope thats ok.)

Tested it on Android, needs Testing on other Platforms.

Best Regards Eric

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -31,11 +31,11 @@ import '../services/translation_service.dart';
 import '../services/notification_service.dart';
 import 'meshcore_connector_usb.dart';
 import 'meshcore_connector_tcp.dart';
+import '../services/contact_discovery_service.dart';
 import '../storage/channel_message_store.dart';
 import '../storage/channel_order_store.dart';
 import '../storage/channel_settings_store.dart';
 import '../storage/channel_store.dart';
-import '../storage/contact_discovery_store.dart';
 import '../storage/contact_settings_store.dart';
 import '../storage/contact_store.dart';
 import '../storage/message_store.dart';
@@ -144,7 +144,6 @@ class MeshCoreConnector extends ChangeNotifier {
   final List<ScanResult> _scanResults = [];
   final List<ScanResult> _linuxSystemScanResults = [];
   final List<Contact> _contacts = [];
-  final List<Contact> _discoveredContacts = [];
   final List<Channel> _channels = [];
   final Map<String, List<Message>> _conversations = {};
   final Map<int, List<ChannelMessage>> _channelMessages = {};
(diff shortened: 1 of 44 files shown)


lib/connector/meshcore_connector.dart
```diff
@@ -356,6 +356,26 @@ class MeshCoreConnector extends ChangeNotifier {
     return List.unmodifiable(_discoveredContacts);
   }
 
+  String exportDiscoveredContactsJson() {
```
446564: I don't think discovered contacts are part of the connector as they are persisted in app only

446564: I would opt for a discovered contacts service or just part of the discovered contacts storage

446564 (changes requested):
I will test today on a few devices

446564:
after the changes feel free to amend your previous commit and force push

446564:
seems to be working well on linux and android, the only issue which is possibly unrelated is that on android you can't save the file. only share to apps.

ericszimmermann:
> seems to be working well on linux and android, the only issue which is possibly unrelated is that on android you can't save the file. only share to apps.

Yes you would have to share to a file manager. I tried to make it to save as file on android, but I always ended up with only storing it in the not accessible  app storage, where it also will be deleted on app uninstall.
Im a C Firmware programmer and am not that familiar with dart and android. If you have a better way to save a file, please let me know.

lib/connector/meshcore_connector.dart
ericszimmermann: I wanted to write as less change as possible and the there was already the connection for "import" newly discovered messages from companion into the store.

446564:
> > seems to be working well on linux and android, the only issue which is possibly unrelated is that on android you can't save the file. only share to apps.
> 
> Yes you would have to share to a file manager. I tried to make it to save as file on android, but I always ended up with only storing it in the not accessible app storage, where it also will be deleted on app uninstall. Im a C Firmware programmer and am not that familiar with dart and android. If you have a better way to save a file, please let me know.

I think this is the same behavior with contacts export so we can always add other options later for both.

lib/connector/meshcore_connector.dart
446564: these are not on the companion however, but in the app. so I think the discovered contacts storage is a better location.

446564:
sorry what I meant was, instead of adding the import and export for discovered contacts into the meshcore_connector, that it should just have been added to the existing contact_discovery_store.

ericszimmermann:
> sorry what I meant was, instead of adding the import and export for discovered contacts into the meshcore_connector, that it should just have been added to the existing contact_discovery_store.

I don't know how to make this going to work, because I need the instantiated object, which is only available in the connector.
I have tried to keep the import/export functions small and to transfer all functions into the contact_discovery_store in the first commit, but I need the reference to the instantiated _discoveredContacts store. (For import _knownContactKeys is also directly available in connector, as are the Listeners.)

We have now two implementation my first commit, which has the least impact and the one where it is as its own service.
I still prefer the first variant.

Sincerely Eric

ericszimmermann:
Should I make a new Pull-Request with a Cherry-Pick of the first commit onto the current dev branch?

446564:
> Should I make a new Pull-Request with a Cherry-Pick of the first commit onto the current dev branch?

whatever is easier for you is fine.

ericszimmermann:
See PR #436 for cherrypick of first variant.

PR #378, opened 2026-04-14
added timestamping on a message in chats and channels

kkazakov:
I noticed that each message doesn't have a visible timestamp so it was confusing when the message was sent.

Add a simple timestamp "14:33" to the message if today.
Add a "13 Apr, 22:34" if yesterday or earlier.

### DIFF
--- a/lib/l10n/app_bg.arb
+++ b/lib/l10n/app_bg.arb
@@ -804,6 +804,26 @@
   "time_months": "месеци",
   "time_minutes": "минути",
   "time_allTime": "Всичко време",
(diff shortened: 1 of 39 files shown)


446564:
this is part the admittedly poorly labelled Enable message tracing, in options.

446564:
also for reference https://github.com/zjs81/meshcore-open?tab=contributing-ov-file, we generally ask that you please open and issue to discuss before creating a pull request.

kkazakov:
Alright, can this be enabled by default? Not many people know it ...

PR #397, opened 2026-04-21
Cyr2lat compression

HDDen:
Hello! This PR adds an additional compression mode, which is useful when working with, for example, Cyrillic characters. 

The idea is that Cyrillic characters are replaced with similar Latin ones, and compression is achieved through the difference in character size – in UTF-8, Cyrillic characters take up 2 bytes, whilst Latin characters take up 1 byte. In practice, compression of between 18–25% is achieved – the freed-up space can be used either to optimise packet size or to fit a greater amount of information within the limit.

Editing the substitution dictionary is supported – for example, you can extend support to other languages.

Cyr2lat compression is currently only implemented for channels. It is can be enabled for each channel in it's settings, and when enabled, it disables SMAZ.

A counter showing the remaining bytes until the limit is also added; it takes into account the enabled cyr2lat compression and recalculates the remaining limit ‘on the fly’ accordingly.


446564:
Hello and welcome, there are some merge conflicts to be resolved first. As well can you review the [contribution guide](https://github.com/zjs81/meshcore-open?tab=contributing-ov-file) and resubmit, ideally the PR should be discussed before hand and should contain one commit with a clear message of what has changed.

I will mark as draft for now, feel free to reopen when ready or close and open a new PR. If you decide to close first please open an issue to discuss your implementation.

PR #402, opened 2026-04-22
fix: multi ack setting

### LINKED ISSUE #401: multi ACKs
I saw in the privacy settings that there's a thing called Multi ACKs that is, by default, set to 0, with options being 0, 1, or 2. However, in the mesh core documentation, it says multi-ack only has 0 or 1 as options. Are these different?

https://docs.meshcore.io/cli_commands/

Edit: Also, I can't find a good explanation of what this does. My thought is that it sends multiple acknowledgments of a successfully received direct message, but I don't see that explicitly stated anywhere.

446564:
can only be 0 or 1 which is equal to 1 or 2 acks, change to simple toggle

closes #401


446564:
closing because I got sniped

Issue #395, opened 2026-04-20
expanding Location sharing functions


ericszimmermann:
I would like more Location sharing functions.
I know this has some big discussion points, espacially for some point we have to be responsible to airtime. Therefore I ask for a contructive diskussion.

less critical implementation:
- [x] deduplication of markers: marker with same label are getting updated on new receive instead of added as multiple copies. #413 
- [ ] adding Handling for receiving Meshcore-SAR marker
- [ ] sending Locationmarker from Inputbar (with current location) as addition to only sending from map. (default label of the location marker would be the device name so it could be easily updated through the deduplication function, but can be changed tho something like "meetingpoint" or to name some other point of interest)

more critical implementation:
- [ ] when sending location over announcemnts is enabled: timed zero-hop announcement to share current location in the vicinity. (Idea from Meshcore-SAR, intervall as in gps-update-setting, which has a default of 15min.)
- [ ] only for PRIVATE channels and limited automatic sending of own position. (Like whatsapp live sharing: For configurable but limited amount of time, sending a locationmarker with "default defice name label" in an gps-update-setting interval. The sending would be only once, without any retry. Maybe also only if the position has changed and/or a region scope is set as soon as #379 is implemented)

All these changes would be in multiple Pull-Requests, some as reimplementation of #346.

zjs81:
PR #413 ('Improve SharedMarker handling', Alpha9) addresses one part of this — SharedMarker deduplication, path history, and re-fetching from chat. The broader location-sharing improvements requested here are not yet fully implemented, so leaving this open.

Danie10:
I'm also looking for limited automated location sending (with a mimimum acceptable period) and some smart position functiionality where it slows any sending when position has not chnaged enough.

Seeing there are a few ideas listed here, can I also add that there is a toggle or slider for location precision? When one is hiking etc you'd probably want exact precision, but when at home and similar areas I'd like to be able to set it with precision of 200m or 300m. This hides my location but still helps for line of sight, etc planning and showing nodes on the map.

ericszimmermann:
Location precision is at the moment not possible, since Location is Companion-FW spezific. But there was a Mention for maybe overwriting periodically from phone. There it would be possible. But that would be Issue #451

Danie10:
[...]

PR #413, opened 2026-04-25
Improve SharedMarker handling

ericszimmermann:
Hello

I have Improved the SharedMarker handling through following points:
- Deduplication of SharedMarkers: If a SharedMarker from the same Source with same Label is received it is handled as the same SharedMarker.
(In the previous Implementation you would get a lot of SharedMarkers overlapping each other.)
- SharedMarkers build History (when recogized as same) to show its path.
- If a SharedMarker is deleted: You can get it back again through tapping on the poi message in chat or channel.

This is one point from #395

As second commit I did a little bit of Housekeeping:
I recognized, that there is in map_screen.dart, chat_screen.dart and channel_chat_screen the same code. Therefore I consolidated these into one `parseMarkerText(String text)` function in map_screen.dart

Sincerely Eric

### DIFF
--- a/lib/l10n/app_bg.arb
+++ b/lib/l10n/app_bg.arb
@@ -682,6 +682,7 @@
   "map_showSharedMarkers": "Покажи споделени маркери",
   "map_lastSeenTime": "Последна видяна дата",
   "map_sharedPin": "Споделено копие",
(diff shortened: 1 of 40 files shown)


zjs81:
Thank you I will get it merged

Issue #73, opened 2026-01-21
[UI/UX] Add unread message marker line


446564:
When heading to a channel or chat with unread messages it's hard to tell where they begin.

ericszimmermann:
And when opening the channel scroll to this line and not to the End/newest one.
Is Related to #43

zjs81:
Implemented in Alpha9 via PR #414 ('Improve work with unread messages'), which adds an unread message separator line, a 'Mark as Unread' action, and jump-to-oldest-unread functionality — directly addressing what was requested here. Closing as resolved.

PR #414, opened 2026-04-26
Improve work with unread messages

Diadlo:
1. Previously setting "Jump to oldest unread" didn't work. It's fixed in a separate commit
2. Added line separator for unread messages and "Mark as Unread" button. I didn't manage to implement universal solution because messages are rendered lazily and we can't calculate the exact place where we need to jump. So just calculate based on the number of message * average message height. Not always great, but covers most of the cases.

Related: #73

Demo:
meshcore-unread-demo


zjs81:
@codex review

zjs81:
Thank you I'll get it merged

PR #416, opened 2026-04-29
Add desktop map controls

wel97459:
This adds controls to zoom in and out in the UX that has a map.
It's only available on desktop since the phone interface is well-rounded for map control.


446564 (approved):
Looks good, working well. Works on mobile normally still.

The only thing I would change would be to add scroll by mouse wheel, feels odd to try try scroll by habit and have the map pan.

(we should also dedupe this map stuff one day lol)

PR #420, opened 2026-05-03
Pin favorited contacts to the top of the contacts list

ChristofferRS:
When you have a large number of contacts, it's usually nice to have favorited contacts appear at the top of the contacts page. You can already filter the list to show only favorites, but having your personal repeater (or similar) always pinned at the top is a small usability win.

## Changes

Favorited contacts are grouped at the top under a "Favorites" header, with the rest under "All Contacts".
When there are no favorites, the list renders exactly as before — no headers.
Existing sort, filter, search, and group selection all still apply within each section.

Section headers are hardcoded English for now. But i am happy to localize via context.l10n in this PR or as a follow-up.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -264,6 +264,8 @@
   "appSettings_languageUk": "Українська",
   "appSettings_enableMessageTracing": "Enable Message Tracing",
   "appSettings_enableMessageTracingSubtitle": "Show detailed routing and timing metadata for messages",
+  "appSettings_enableFavoritesSection": "Separate Favorites Section",
+  "appSettings_enableFavoritesSectionSubtitle": "Show separate favorites section at top of contacts",
   "appSettings_notifications": "Notifications",
   "appSettings_enableNotifications": "Enable Notifications",
   "appSettings_enableNotificationsSubtitle": "Receive notifications for messages and adverts",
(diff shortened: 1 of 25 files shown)


446564:
Thank you for helping out! Can you please review the contribution guide? Features should be discussed before submission.

This should be a configurable setting and include localizations. I will mark it as draft, please change it back when ready.

ChristofferRS:
> Thank you for helping out! Can you please review the contribution guide? Features should be discussed before submission.

@446564 i apologies. I was a bit to quick to make the PR 😅 Would you like me to open an issue now or just implement the features as you proposed then below:

> This should be a configurable setting and include localizations. I will mark it as draft, please change it back when ready.

446564:
you can just go ahead with the changes, and checkout the doc re: formatting. Basically just run `dart format .` before committing. If you need to add any strings, add them to `lib/l10n/app_en.arb` and after run `flutter gen-l10n -v`, then the will be available to use from `context.l10n.my_string` for example.

ChristofferRS:
@446564 I have addressed your feedback:

- Added a setting to enable/disable the feature, defaulting to disabled
- Added all strings to l10n in what I hope is a sensible location
- Ran `dart format .`
- Squashed to a single commit as per the contribution guide
- Tested several combinations: different groups, with and without favorites, with the setting enabled and disabled

Screenshots below for reference:


## Settigns Page

settings

## Default behavior (no favorites)

default

## Favorites section with groups

group

## Favorites section on contacts screen all

All-Page

446564 (dismissed):
nice work

i could go either way on the location of the setting, if that's as is in app settings or a checkbox in the contact list filter menu

446564:
just check the failed dart analyze for nit pick about fence blocks with if statements

ChristofferRS:
> nice work
> 
> i could go either way on the location of the setting, if that's as is in app settings or a checkbox in the contact list filter menu

I was also on the fence on where to put it but in the end i decided that app settings might be marginaly more intuitive, but honestly i am also not sure.  I am happy to move if needed.


I think the analyze issue should be fixed now, with a bit of additional git gymnastics to resolve a merge conflict and keep it at 1 commit :slightly_smiling_face:

446564 (approved):

Issue #312, opened 2026-03-25
Feature Request: Channel Utilisation


holdit:
In meshtastic, it's usually easy to see the channel utilisation. In meshcore, not so much.

Since meshcore is starting to get a bit busy in some places, it would be nice if we could see this info somewhere, especially for repeaters.

Not sure if it's possible, but posting here for your consideration.

Thanks!

446564:
easy enough for repeaters as we have uptime, tx time and rx time.

for companion it might need a firmware change, or only work after being connected to the device for a while and have to reset on disconnect.

holdit:
MeshCore SAR ( https://github.com/dz0ny/meshcore-sar/ ) expands companion stats a bit. No channel util, but gives us an idea of what's going on.

Nice UI in general, very polished, but not exactly ideal to use as a normal app as it requires location and it's very focused on search and rescue/off-grid group stuff. Could be a good source of inspiration for you guys.

<img width="1080" height="1638" alt="Image" src="https://github.com/user-attachments/assets/a1e85cec-a89f-4e2e-8dbe-e9b8148aee29" />

zjs81:
Implemented in Alpha9 via PR #424 ('basic repeater chan util') — the repeater status screen now displays a channel utilisation percentage calculated from TX/RX air-time and uptime data. Closing as resolved.

PR #424, opened 2026-05-05
basic repeater chan util

446564:
just repeater stats for now

total channel utilization over uptime i.e. util =
(rxsecs+txsecs) / upsecs

partial #312


446564:
companion can be a follow up as the connector to fetch stats needs updating, maybe we can use a rolling live window instead like noise floor and show a graph.

Issue #418, opened 2026-05-01
Remove channel subtitles in channel list


sethoscope:
The channel list includes a line of text for each channel, saying what kind of channel it is: "Public channel", "Hashtag channel", or "Private channel". This adds a lot of visual clutter, obscuring the channel names. The UI is clearer and more compact without the subtitles.

Here are screenshots with and without the subtitles for comparison.

<img width="250" alt="with subtitles" src="https://github.com/user-attachments/assets/e711db73-5446-457d-9d60-cb7cb6fc39ca" />

<img width="250" alt="without subtitles" src="https://github.com/user-attachments/assets/e1719f7f-4eb3-441f-b1d2-94400c80f995" />

446564:
Ya I agree, this is redundant and cluttered. We might have to keep something for community channels.

sethoscope:
Community channels have their own icon though. What would you need to put in the extra line of text?

446564:
> Community channels have their own icon though. What would you need to put in the extra line of text?

ya probably nothing as we have tooltips as well.

sethoscope:
I'd be happy to turn this into a PR if you like what you see. Or to change it if you want something different. The screenshot is from what I'm running on my phone.

446564:
I would PR it, looks cleaner.

sethoscope:
I noticed there's one other place in the UI that uses one of these subtitles: in the map view, if you create a marker to share, it gives you a list of contacts and channels you can share it with. Oddly, only the Public channels get a subtitle, and the other types don't. I'm inclined to remove it here too, but also I think it's less of a problem and I don't feel strongly about it.

<img width="300" alt="map view point share" src="https://github.com/user-attachments/assets/7a2ce7f5-e830-48fa-a4c2-7174d1639373" />

Also, how do you feel about me deleting the localization strings for these subtitles now that we're not using them for anything? Keep them just in case or tidy things up and don't maintain stuff we don't use?

446564:
I agree with both, remove it in the send to channel dialog and remove any unused localizations

sethoscope:
Fixed with 08ac60a408c1aca3275b31d1372d2ad65ef75272

PR #428, opened 2026-05-06
Remove channel subtitles from UI

sethoscope:
Per issue #418, this commmit removes channel subtitles from the channel list and from the map screen (deep in the marker sharing). This reduces visual clutter and allows for more compact lists, and the type of channel is already indicated by the leading icon.

The subtitles simply said "Public channel", "Hashtag channel", or "Private channel".

We also remove the relevant localization strings.


446564:
I feel like they need a little more padding top and bottom, in the cards. looks squished now

sethoscope:
I agree, it was a bit cramped. I added more/different padding.
I rebased & squashed so you still get just one commit.

screenshot

446564 (approved):
looks much cleaner

PR #429, opened 2026-05-06
fix: lat/lon double-write in buildUpdateContactPathFrame

### LINKED ISSUE #427: bug: buildUpdateContactPathFrame writes lat/lon twice, corrupting the optional timestamp tail
## Bug

`buildUpdateContactPathFrame` in `lib/connector/meshcore_protocol.dart:723-747` emits two consecutive 8-byte position blocks instead of one. The frame ends up 8 bytes longer than the documented layout, and any caller that also passes `lastModified` writes the timestamp 8 bytes deeper than the firmware expects.

## Frame layout per the doc comment

The function is documented (line 695) as producing:

So the optional tail is 8 bytes of position + optional 4-byte timestamp — 12 bytes max.

## What it actually emits

Two back-to-back blocks both write position data:

```dart
// Block 1 (lines 723-736) — runs unconditionally
if ((lat == null || lon == null) && lastModified != null) {
  writer.writeInt32LE(0);
  writer.writeInt32LE(0);
} else {
  final latitude = lat ?? 0.0;
  writer.writeInt32LE((latitude * 1e6).round());
  final longitude = lon ?? 0.0;
  writer.writeInt32LE((longitude * 1e6).round());
}

// Block 2 (lines 738-747) — runs whenever location OR lastModified is set
final hasLocation = lat != null && lon != null;
if (hasLocation || lastModified != null) {
  writer.writeInt32LE(hasLocation ? (lat * 1e6).round() : 0);
  writer.writeInt32LE(hasLocation ? (lon * 1e6).round() : 0);
  if (lastModified != null) {
    final lastModifiedTimestamp = lastModified.millisecondsSinceEpoch ~/ 1000;
    writer.writeUInt32LE(lastModifiedTimestamp);
  }
}
```

Block 1 runs unconditionally and emits 8 bytes. Block 2 then adds another 8 bytes whenever location or `lastModified` is set, plus the optional 4-byte timestamp.

For `buildUpdateContactPathFrame(pk, path, n, lat: 49.1, lon: -123.1, lastModified: now)`:

| Expected | Actual |
|---|---|
| `[lat x4][lon x4][ts x4]` (12 bytes) | `[lat x4][lon x4][lat x4][lon x4][ts x4]` (20 bytes) |

The firmware parses the second `lat` field as the timestamp, getting wildly wrong values.

## Reproduction

```dart
final frame = buildUpdateContactPathFrame(
  Uint8List(32), Uint8List(0), 0,
  lat: 49.0, lon: -123.0, lastModified: DateTime.now(),
);
// Base frame is 136 bytes (cmd + pubKey + type + flags + pathLen + path + name + timestamp)
expect(frame.length, 148); // 136 + 8 lat/lon + 4 timestamp
// Actual: 156 (= 136 + 16 lat/lon + 4 timestamp)
```

## Proposed fix

Delete block 1; keep only block 2. Block 2 is structurally correct: skips the optional tail when neither location nor `lastModified` is set, zero-fills position slots when only `lastModified` is present, appends the optional timestamp when applicable.

I have a fix + 5-test suite (covering all four optional-tail combinations) ready in a downstream fork. Happy to open a PR if you'd like — let me know.

## Context

Caught while adding wire-protocol tests to a downstream fork. There were no tests covering this function previously, which is why the bug wasn't surfaced earlier.

Affected user-facing path: MeshCoreConnector.importDiscoveredContact() at line 3085 is the only caller that passes all three optional args (lat, lon, lastModified). That means every imported discovered contact gets a corrupted timestamp on the radio — probably manifesting as wildly wrong "last seen" values in the UI.

kingult:
Fixes #427.

`buildUpdateContactPathFrame` emitted two consecutive 8-byte position blocks instead of one, producing a frame 8 bytes longer than the documented layout. When a caller passed `lastModified`, the firmware parsed the duplicated second `lat` field as the timestamp, giving wildly wrong "last seen" values on imported contacts.

## Change

Delete the unconditional first block; keep the conditional block. The conditional block is structurally correct: it skips the optional tail when neither location nor `lastModified` is set, zero-fills position slots when only `lastModified` is present, and appends the optional 4-byte timestamp when applicable.

446564 confirmed this approach in the issue thread — "Keeping block 2 was going to be my fix here."

## Test coverage

Adds `test/connector/build_update_contact_path_frame_test.dart` with five cases:

1. Neither location nor `lastModified` provided → no optional tail (exactly `baseFrameLength` bytes).
2. Location only → 8-byte lat/lon tail.
3. Both location and `lastModified` → 8 + 4 = 12-byte tail.
4. `lastModified` only → 8 zero bytes for the position slot + 4-byte timestamp, with byte-level zero-check on the position slot to guard against future regressions writing garbage there.
5. Positive lat/lon encoding sanity: little-endian, fixed-point ×1e6.

## Verification

- `dart format` clean
- `flutter analyze --no-fatal-infos` clean
- `flutter test test/connector/build_update_contact_path_frame_test.dart` — all five tests pass

### DIFF
--- a/lib/connector/meshcore_protocol.dart
+++ b/lib/connector/meshcore_protocol.dart
@@ -720,27 +720,16 @@ Uint8List buildUpdateContactPathFrame(
   final timestamp = DateTime.now().millisecondsSinceEpoch ~/ 1000;
   writer.writeUInt32LE(timestamp);
 
-  if ((lat == null || lon == null) && lastModified != null) {
-    // If lat/lon not provided, write zeros
-    writer.writeInt32LE(0);
-    writer.writeInt32LE(0);
-  } else {
-    // Latitude and Longitude are expected in degrees, convert to int by multiplying by 1e6
-    // Latitude
-    final latitude = lat ?? 0.0;
-    writer.writeInt32LE((latitude * 1e6).round());
-
-    // Longitude
-    final longitude = lon ?? 0.0;
-    writer.writeInt32LE((longitude * 1e6).round());
-  }
-
+  // Optional [Lat x4, Lon x4][timestamp x4] tail per the doc comment above.
+  // Emit 8 bytes of position (zero-filled when only lastModified is provided)
+  // followed by an optional 4-byte timestamp. Earlier code emitted the
+  // position block twice, which corrupted the tail and caused the firmware
+  // to parse the second lat as the timestamp. See #427.
   final hasLocation = lat != null && lon != null;
   if (hasLocation || lastModified != null) {
     writer.writeInt32LE(hasLocation ? (lat * 1e6).round() : 0);
     writer.writeInt32LE(hasLocation ? (lon * 1e6).round() : 0);
     if (lastModified != null) {
-      // Last modified
       final lastModifiedTimestamp = lastModified.millisecondsSinceEpoch ~/ 1000;
       writer.writeUInt32LE(lastModifiedTimestamp);
     }
--- a/test/connector/build_update_contact_path_frame_test.dart
+++ b/test/connector/build_update_contact_path_frame_test.dart
@@ -1,0 +1,105 @@
+import 'dart:typed_data';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:meshcore_open/connector/meshcore_protocol.dart';
+
+void main() {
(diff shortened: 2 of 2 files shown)


446564 (approved):
Thanks for the catch and the tests 😎

PR #436, opened 2026-05-10
import export function for discovered contacts Nr.5

ericszimmermann:
Made the changes for importing and exporting Discovered contacts as JSON files.

This is according to Issue https://github.com/zjs81/meshcore-open/issues/352 (and based on now obsolete https://github.com/zjs81/meshcore-open/pull/373 and https://github.com/zjs81/meshcore-open/pull/298).
This is a cherry-pick of the first variant in #373.

I also cleaned up the code a bit, and this is the isolated import/export functionality.

There was just on thing. there was a contactfield "wasPulled" which seemed to should have the same functionality as the existing "isActive" flag, but was not used anywhere, therefore I removed it. (I hope thats ok.)

Tested it on Android and verified https://github.com/zjs81/meshcore-open/pull/373#issuecomment-4290809383

Best Regards Eric


ericszimmermann:
I keep this updated since first Pull-Request for it in Alpha7, because I still use it to collect path hop information
and tranfer them between all my testbuilds.

PR #443, opened 2026-05-11
Add "NRF52" as a device name prefix

Maxb0tbeep:
Hello, for some reason, my T-Echo decides to rename itself to "NRF52 DK", which breaks the app from being able to discover it. I think this might be caused by the node name being too long to fit in a bluetooth device name, but this is fixes it anyways. I tested it on my phone and it resolves the issue.


zjs81:
Thanks for submitting this. It looks good I will get it merged.

Issue #432, opened 2026-05-08
Channel chat shows # for private channels


sethoscope:
At the top of the channel chat screen, there's an icon to the left of the channel name. It's the public icon for the public channel, and the hashtag icon for both hashtag and private channels. For private channels, it should use the lock icon to be consistent with the channels list.

Example of how it is now:

<img width="300" alt="channels" src="https://github.com/user-attachments/assets/675892e7-2a40-4808-a23c-feacca69d3a9" />
<img width="300" alt="channel chat" src="https://github.com/user-attachments/assets/4a7418d4-4351-461b-bb11-97a9ae5a0b9e" />

Example of how it should be:

<img width="300" alt="channel chat" src="https://github.com/user-attachments/assets/fca1da90-3c79-44e5-9105-fb5f3ee1678f" />

Easy fix, already written, happy to submit a PR.

446564:
please do :)

PR #444, opened 2026-05-11
Use correct channel icons in channel chat screen

sethoscope:
At the top of the channel chat screen is an icon, indicating the channel type.

Previously, the public icon was used correctly, but the hashtag icon was used for all other types.

Now, consistent with the channels screen, we use the lock icon for private channels, and the composite icons for community public & community hashtag types.

Closes issue #432.

### DIFF
--- a/lib/screens/channel_chat_screen.dart
+++ b/lib/screens/channel_chat_screen.dart
@@ -8,6 +8,8 @@ import 'package:intl/intl.dart';
 import 'package:provider/provider.dart';
 
 import '../connector/meshcore_connector.dart';
+import '../models/community.dart';
+import '../storage/community_store.dart';
 import '../utils/platform_info.dart';
 import '../helpers/chat_scroll_controller.dart';
 import '../connector/meshcore_protocol.dart';
@@ -56,8 +58,11 @@ class _ChannelChatScreenState extends State<ChannelChatScreen> {
   final ChatScrollController _scrollController = ChatScrollController();
   final FocusNode _textFieldFocusNode = FocusNode();
   ChannelMessage? _replyingToMessage;
+  final CommunityStore _communityStore = CommunityStore();
+  final CommunityPskIndex _communityIndex = CommunityPskIndex();
   final Map<String, GlobalKey> _messageKeys = {};
   bool _isLoadingOlder = false;
+  bool _communitiesLoaded = false;
 
   MeshCoreConnector? _connector;
   DateTime? _lastChannelSendAt;
@@ -81,6 +86,7 @@ class _ChannelChatScreenState extends State<ChannelChatScreen> {
       final idx = widget.channel.index;
       final unread = widget.initialUnreadCount;
       final messages = connector.getChannelMessages(widget.channel);
+      _loadCommunities();
       ChannelMessage? anchor;
       if (unread > 0) {
         anchor = _findOldestUnreadChannelAnchor(messages, unread);
@@ -105,6 +111,19 @@ class _ChannelChatScreenState extends State<ChannelChatScreen> {
         });
       }
     });
+  }
+
+  // TODO: Reload communities when returning from another screen
+  Future<void> _loadCommunities() async {
+    final connector = context.read<MeshCoreConnector>();
+    _communityStore.setPublicKeyHex = connector.selfPublicKeyHex;
+    final communities = await _communityStore.loadCommunities();
+    if (mounted) {
+      setState(() {
+        _communityIndex.initialize(communities);
+        _communitiesLoaded = true;
+      });
+    }
   }
 
   ChannelMessage? _findOldestUnreadChannelAnchor(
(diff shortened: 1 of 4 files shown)


sethoscope:
Just a note... Originally I only thought to fix the private channel icon, and that was a 2-line change. But as I dug in more, I finally understood what community channels are and how they work, and those icons were wrong too. 

Fixing that was much more involved, because it's not easy to tell whether a channel is part of a community. I ended up duplicating some things from channels_screen.dart. I tried a few times to refactor that to make it more reusable, but I didn't come up with anything I liked, and I decided this was better for now.

lib/screens/channel_chat_screen.dart
```diff
@@ -107,6 +112,17 @@ class _ChannelChatScreenState extends State<ChannelChatScreen> {
     });
   }
 
+  Future<void> _loadCommunities() async {
+    final connector = context.read<MeshCoreConnector>();
+    _communityStore.setPublicKeyHex = connector.selfPublicKeyHex;
+    final communities = await _communityStore.loadCommunities();
```
zjs81, line 120: Fire-and-forget async load means a community channel renders as `Icons.lock` (the `private` default) on the first frame, then swaps once communities load. On slower devices this flickers visibly.

Two options:
- Render a neutral placeholder (or hide the icon) until `_communities` has loaded at least once, or
- Make `CommunityStore.loadCommunities()` return a synchronously-cached list when one is available.

Also note: `_communities` never refreshes while this screen is open, so edits made elsewhere won't be reflected until the screen is recreated — acceptable for now but worth a one-line comment if you don't want to wire it to the store.

lib/screens/channel_chat_screen.dart
```diff
@@ -193,16 +209,82 @@ class _ChannelChatScreenState extends State<ChannelChatScreen> {
...
+        }
+      }
+    }
+    if (channel.isPublicChannel) {
+      return ChannelType.public;
+    } else if (channel.name.startsWith('#')) {
+      return ChannelType.hashtag;
+    }
+    return ChannelType.private;
+  }
```
zjs81: **Please dedupe this with `channels_screen.dart` in this PR rather than as a follow-up.**

`channels_screen.dart:378-409` already classifies channels with the same logic via `_getCommunityForChannel` / `_isCommunityPublicChannel`. With the new `ChannelType` enum living in the model, the natural home for this is a static helper alongside it, e.g.:

```dart
// lib/models/channel.dart
enum ChannelType { public, private, hashtag, communityPublic, communityHashtag }

extension ChannelTypeClassify on Channel {
  ChannelType classify(Iterable<Community> communities) {
    for (final community in communities) {
      if (pskHex == Channel.formatPskHex(community.deriveCommunityPublicPsk())) {
        return ChannelType.communityPublic;
      }
      for (final hashtag in community.hashtagChannels) {
        if (pskHex == Channel.formatPskHex(community.deriveCommunityHashtagPsk(hashtag))) {
          return ChannelType.communityHashtag;
        }
      }
    }
    if (isPublicChannel) return ChannelType.public;
    if (name.startsWith('#')) return ChannelType.hashtag;
    return ChannelType.private;
  }
}
```

Then `channel_chat_screen.dart` calls `widget.channel.classify(_communities)` and `channels_screen.dart` is refactored to use the same call instead of its ad-hoc `_getCommunityForChannel` / `_isCommunityPublicChannel`. That keeps the two screens from drifting as new channel types are added.

Also rename to `_getChannelType` / drop entirely once this is extracted.

lib/screens/channel_chat_screen.dart
```diff
@@ -193,16 +209,82 @@ class _ChannelChatScreenState extends State<ChannelChatScreen> {
...
+        icon = Icons.groups;
+      case ChannelType.communityHashtag:
+        icon = Icons.tag;
+      case ChannelType.public:
+        icon = Icons.public;
+      case ChannelType.hashtag:
+        icon = Icons.tag;
+      default:
+        icon = Icons.lock;
+    }
```
zjs81, line 234: Two changes here:

1. **Replace `default:` with `case ChannelType.private:`.** With an explicit case, Dart's switch exhaustiveness check fires if a new `ChannelType` is added later — `default:` silently swallows it.
2. **Rename `channelIcon` → `_channelIcon`** to match the file's convention for state-private methods.

lib/screens/channel_chat_screen.dart
```diff
@@ -193,16 +209,82 @@ class _ChannelChatScreenState extends State<ChannelChatScreen> {
...
+    return Stack(
+      // Use a stack in case we want to add community annotation
+      children: [
+        Padding(
+          padding: const EdgeInsets.symmetric(vertical: 3),
+          child: Icon(icon),
```
zjs81: Visible regression: the previous AppBar icon was `Icon(..., size: 20)`. This `Icon(icon)` falls back to the AppBar default (24), so the title-row icon is now noticeably larger than before.

```suggestion
          child: Icon(icon, size: 20),
```

If you keep the size, also scale the community badge `Container` (currently 14×14) proportionally so the annotation doesn't overflow the smaller icon.

lib/screens/channel_chat_screen.dart
```diff
@@ -193,16 +209,82 @@ class _ChannelChatScreenState extends State<ChannelChatScreen> {
...
+        icon = Icons.tag;
+      case ChannelType.public:
+        icon = Icons.public;
+      case ChannelType.hashtag:
+        icon = Icons.tag;
+      default:
+        icon = Icons.lock;
+    }
+    return Stack(
+      // Use a stack in case we want to add community annotation
```
zjs81: Nit: this comment is misleading — the `Stack` is already used for the community-badge annotation a few lines below, not "in case we want to add" one. Either remove or reword to describe what it actually does.

zjs81 (changes requested):
Thanks for the fix — the classification matches `channels_screen.dart` and renders correctly. A few things to address before merge:

1. **Dedupe the classification with `channels_screen.dart`.** The same 5-way logic now lives in two places (`channels_screen._getCommunityForChannel`/`_isCommunityPublicChannel` and the new `getChannelType` here). Please pull this into a shared helper in this PR — e.g. a static `ChannelType.classify(Channel channel, Iterable<Community> communities)` next to the new enum in `lib/models/channel.dart` — and refactor `channels_screen.dart` to use it too. The enum is already in the model, so the logic should live there as well; otherwise the two screens will drift.
2. **Restore `size: 20` on the AppBar icon** — current code falls back to the default (24), which is a visible regression versus the previous `Icon(..., size: 20)`.
3. **Replace `default:` with explicit `case ChannelType.private:`** so the Dart 3 switch is exhaustive and the analyzer warns when a new variant is added.
4. **Prefix new helpers with `_`** (`_getChannelType`, `_channelIcon`) — they're state-internal and the rest of the file uses that convention.
5. **First-frame flicker** — `_loadCommunities()` is fired async after the first frame, so community channels render as `Icons.lock` (private) until it resolves. Worth a placeholder or a synchronous cached read; see inline.
6. Minor: the `// Use a stack in case we want to add community annotation` comment is misleading — the Stack is already doing that.

No test changes needed.

lib/screens/channel_chat_screen.dart
sethoscope: I meant "in case we want to add one" because the inclusion of the badge is conditional, not for future development. But the comment clearly does more harm than good, so out it goes.

lib/screens/channel_chat_screen.dart
sethoscope: Thanks. I'm new to Flutter & Dart, so I'm still getting up to speed on a lot of this. 

Does channels_screen have the same flicker problem? It looks like the only difference between the two is that here _loadCommunities() is run from SchedulerBinding.instance.addPostFrameCallback and in channels_list it's from WidgetsBinding.

I regret to say I don't understand the second option.

Re the refresh note: I think I get it. We only load communities the first time, and if we navigate back here after changing them, the list will be stale? 20 minutes of reading about this stuff has mainly taught me that there are good idiomatic ways of managing this in Flutter, but I don't yet know what they are.

I force pushed what I have so far, which fixes all the issues you pointed out except this one.

sethoscope:
@zjs81 
Ok, I think I've fixed all the problems you pointed out, other than refreshing the community list in channel_chat_screen after it's updated elsewhere.

I didn't observe the flickering problem, so I'm taking my fix on faith. Probably I should get set up with some virtual test devices.

zjs81:
Thanks I'll review it again.

lib/screens/channel_chat_screen.dart
zjs81: I'll do some testing I've noticed this before on my phone when I had lots of channels and I put them in custom positions.

PR #453, opened 2026-05-17
Room-server: fixed first message letters trim

HDDen:
On room-servers, the first 4 characters of all incoming messages are truncated. Also, sender names was truncated. With this fix, they are now displayed in full again.

How to reproduce: Go to a room server, receive new messages, and check them.

Next, apply the fix and check again. New messages should now appear in full.

This fix was created using Codex: I took the current implementation in meshcore_open and compared it with the implementations from the firmware and meshcore_py. The fix involves aligning the mechanism with the latter two sources.

<details>
  <summary>Screenshot</summary>
  
  room-server
  
</details>


zjs81:
@codex review

Issue #437, opened 2026-05-10
[accessibility] Dark/Satellite Map


VioletStorm:
Absolutely gorgeous little app, but the map tab is a flashbang (Android), really rough with light sensitivity, a dark theme option or even just forced satellite mode would help so much!

PR #461, opened 2026-05-23
Alternative Maps

ericszimmermann:
I looked to implement Issue #437.

To provide other map styles, especially a dark map.
For that I implemented a [Stadiamap](https://stadiamaps.com/) Endpoint.
With the free account you can download 200'000 tiles per month of various styles.
You will have to create an API-Key and insert it into the app.

I did not find a fully free provider without registration.
The Standard OpenStreetMap tile server is still supported for direct view.

Best Regards Eric


shortwavesurfer2009:
Wouldn't this cause the label depends on non-free network services to be added if it was added to F-droid?

ericszimmermann:
If it were on f-droid probably yes.
But it does not depend on it, the alternative maps are just an option.
Unfortunately bulk downloads are not allowed for free OSM tileserver.
Do you know of a fully free tile server?

shortwavesurfer2009:
I am not aware of any. OSM doesnt provide dark tiles?

Edit: I wonder how the meshtastic map gets its dark tiles. It says they are provided by OSM.

https://meshmap.net/

ericszimmermann:
No I have not found any mention of dark tiles in the official openstreetmap tile server and in terms of service prohibits bulk downloads.

ericszimmermann:
meshmap.net just inverts the color with
```
  body.dark {
    filter: invert(1) hue-rotate(180deg) brightness(1.33);
  }
```
Implemented similar function here: Invert tile png.

shortwavesurfer2009:
I wanted to test this so i built an apk from source for the very first time using your branch.

Screenshot_20260524-093715_meshcore_open
Screenshot_20260524-093935_meshcore_open

shortwavesurfer2009:
Would it be possible to make the map switch between OSM and OSM dark depending on your system theme by default?

Edit: i know the OSM ones are foss, what about those others?

ericszimmermann:
I would have to have a look at the theme switching code: then I could maybe do the switches if you are at the OSM maps.

The others are Stadiamaps. I looked at some map tiles provider and liked them best.
You can get a free subscription there, which allows you to download up to 200'000 tiles.
Of corse its a dependancy, but everyone is free to decide to use it or not.
For a lot of provider you would have to register as the app provider and pay for all the users, whereas here the user has to register themselves and get the API to enter into the app. So no these others are not FOSS.

But I also wanted to keep the offline cache mode. Which is a function I really like.
Just upon my research for the dark map style I found that bulk downloads are prohibited in the service agreements of OSM. So I deaktivated this function for the OSM maps. Not that they deactivate the whole access for our app.

Best regards Eric

shortwavesurfer2009:
Just built this to test it out, and it works great. The map was automatically in dark mode, because the app was automatically in dark mode, because the system is in dark mode, and I was never blinded by a white map. Beautiful.

shortwavesurfer2009:
Damn. @zjs81 didn't review this PR before making the new release version.

I will wait for you to update this and build it myself before I update.

ericszimmermann:
The new app theme also includes a new dark map, similar to my "OpenStreetMap Dark" provider. Naturally, they now compete with each other, but I've tried to consolidate them so that I can still offer StadiaMaps integration:
After all, even though it is officially not permitted, bulk downloads from the OSM tile server are still allowed in the current dev-branch. I would hate if they would block all, just for those who use bulk downloads to extensively.

shortwavesurfer2009:
Ah, okay. I didn't see anything about a dark map being integrated, so I didn't want to use the new version until I knew that it would stay dark.

ericszimmermann:
@zjs81 for https://github.com/zjs81/meshcore-open/issues/480#issuecomment-4722917473:

Yes I am finished with updating this PR. It's just one of the test in a not to this PR relatet part has not successfully finished.
I also included access to my free subscription with a secondary key.

Sincerely Eric

PR #484, opened 2026-06-18
Fix adaptive icon and app label on Android

hamid-elaosta:


hamid-elaosta:
Adaptive Icon was displaying incorrectly on my Pixel 9 Pro XL. Fixed and regenerated icons.

The label was also showing incorrectly, also updated.

zjs81:
thanks

PR #491, opened 2026-06-23
Add Australia (Mid) Radio Preset

km-foss:
This mirrors the config option from the official Meshcore app, and represents the preset now commonly in use in New South Wales.

Being a minor update which mirrors the official app, I have gone straight to PR, but happy to discuss if needed.

For reference, from the Meshcore app:

image

- [x] Docs updated
- [x] `flutter analyze` run
- [x] `dart format` run


zjs81:
Thank you!

Issue #477, opened 2026-06-13
Feature Request: Implement some features from Meshcore Team


MrSurly:
I have a use case (hiking, essentially) where it would be great to keep track of teammate locations, in case people get separated.  

Meshcore Team does this, but their app has stability issues.

Note that I am an embedded developer, but not a Dart/Flutter developer.  I've begun learning, and I would be willing to implement this myself and submit as a PR, but I don't want to work on it if it won't be accepted.

Features:

* Add setting screen for "locations sharing" that lets you select a private channel to share your location:
    * Share location max frequency (i.e. "don't send location more often than X minutes")
    * Share location min frequency (i.e. "share location if it's been more the X minutes")
    * Minimum distance delta for location sharing (i.e. "share location if you've moved a minimum of X meters")
 * Map shows team locations, with an option to only show locations received via private channel

All location stuff is combined;  e.g. location delta is only shared if it's been more than min frequency
    
Nice to have:

* Map screen shows your bearing (via device magnetometer)
* Map screen let's you toggle between "north at top" and "your bearing at top"
* Map screen shows path/tracking of other nodes 

This isn't inclusive of all the Team features, but I feel it's a start.

ericszimmermann:
That is similar to Issue #395, where I began to implement things for it. But I haven't found enough time.
What already exists is when markers are send with same label and flag they update and draw a path on map.
What is missing is a timed resending. I indented to use the configured GPS intervall for a limited configurable amount of time and only in a private channel.
And there is also the proposal to have the location send out with timed zero-hop intervals, which could also be handy.

MrSurly:
Yup, sounds very familiar

MrSurly:
From #395 : configurable but limited amount of time, sending a locationmarker

I like this.  I am concerned about the problem Meshtastic has of people spamming location too much.  I think something where you can set it up to do this for a limited time (even if it's measured in days) is a good idea.

sethoscope:
Pushing telemetry to the mesh is frowned upon, but maybe this could be available only in Off-Grid Repeat mode. That's usually what you'd want for a hiking group anyway.

stepedro83:
It would make sense to have a programmable zero hop advert with the location updated on a regular time or distance. It's a nice tradeoff approach for teams.

ericszimmermann:
I just tested my Implementation for location zero-hop advert and will make the Pull-request #497.

PR #497, opened 2026-07-01
Zero-Hop Location advert

ericszimmermann:
This is an other part of Issue #395 and also a request from #477:

Implementation of automated zero-hop adverts for location sharing, when the option for location in adverts is enabled.
To mitigate spamming, the are some condition for a zero-hop advert is send:
- It will be send in the configured gps interval if,
- location advert and specific setting for this function are enabled and
- you moved about 25m from last location when not in clientRepeat (Hiking) mode.
   (In clientRepeate mode it will always be sent)

To have a reliable gps intervall, the settings is now also stored in the app setting in addition to the device:
- on read of gps settings it will try to read it from device, upon failure it takes the configured app setting or default 900s (15min).
- on write it will write into app settings and try to write it on device. If write fails, the device can have a fix 1s update intervall in the firmware, therefore there is a explicit gps intervall filter in the self info update handler before sending advert.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -216,6 +216,9 @@ class MeshCoreConnector extends ChangeNotifier {
   DateTime _lastRadioRxTime = DateTime.fromMillisecondsSinceEpoch(0);
   DateTime _lastContactMsgRxTime = DateTime.fromMillisecondsSinceEpoch(0);
   DateTime _lastChannelMsgRxTime = DateTime.fromMillisecondsSinceEpoch(0);
+  DateTime _lastZeroHopAdvertAt = DateTime.fromMillisecondsSinceEpoch(0);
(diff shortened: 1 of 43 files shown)


RevelRob:
If clientRepeat is enabled, does it then become a flood advert as opposed to zero hop?

ericszimmermann:
No it is still zero hop.
In clientRepeat mode it is only the movement restriction which is ignored. Because this restriction is there so that the air channel is less overwhelmed with the zerohop signals if there is not new information.
In clientRepeat mode it resends the zerohop advert with every intervall, because we may afford it, since its on its own air channel. It is so, that if you lost connection and get back into reach you will get the new exact location, even if the other group is making a rest or camp.

If its important that it should be flood advert, you will have to open an issue, where it can be discussed and approved.

Best Regards Eric

PR #506, opened 2026-07-23
Add SWHID and Archive badges to README

Warmax356:
Added SWHID and Archive badges to the README.

p.s i am an ambassader of Software heritage and have curated the project to be included into the libreary and provided the SWHID this will help with SBOMS in CRA


zjs81:
Very nice thank you

Issue #500, opened 2026-07-01
meshcore_open is using too much memory


celsoazevedo:
Sometimes my Samsung phone warns me about this app using too much memory:

<img width="953" height="342" alt="Image" src="https://github.com/user-attachments/assets/0188678d-b446-4017-a2a6-0f34ad30f5dd" />

This notification was shown after leaving MeshCore Open running in the background for a few hours.

It has been a long while since I saw this notification for any app, so I'm assuming there's something about MeshCore Open that sometimes triggers it. Perhaps a leak or something that isn't as optimized as it could be, and memory use keeps increasing?

zjs81:
Thank you I will look into this

ericszimmermann:
Well, I never had the problem with memory. Could the current solution to prune the discovered contact list be made configurable?
Or is there an other solution planned? Then the current pruning is has other disadvantages.
For example here I have a path, one time where I have imported my discovered contacts, ...
<img width="1312" height="5930" alt="Image" src="https://github.com/user-attachments/assets/d322e698-743d-4434-8b74-7c8a5043036b" />
and the second after the discovered contact list was pruned through restarting the app:
<img width="1312" height="5930" alt="Image" src="https://github.com/user-attachments/assets/b8350e3e-d49c-4609-8cb2-14e60d00f9c4" />
You will see that there is not much information of the path left.

celsoazevedo:
I have no idea what caused the OS to show that warning and not being a dev, I have no idea on how to fix it, but the official app seems to handle +2000 discovered contacts without difficulties (that I have noticed). I've reinstalled the app since then, but when it happened, I think I had fewer than 400 contacts in MeshCore Open. If it's this, then there seems to be room for improvement.

PR #512, opened 2026-08-03
Setting to enable eviction of discovered contacts

ericszimmermann:
I made the eviction of of discovered contacts configurable in the contact settings.
For those few which have memory problems as in discribed in #500 [meshcore_open is using too much memory](https://github.com/zjs81/meshcore-open/issues/500).
The problem is with this settings you will loose a lot of information for searching the route. (also in Issue)
Because of that I beleave it should be disabled by default.

Sincerely Eric


zjs81:
@codex review

PR #513, opened 2026-08-06
Add self node telemetry screen

Hacuchino-hash:
## Summary

Adds a telemetry view for the locally connected node. The firmware answers self telemetry requests over the companion protocol (the 4 byte CMD_SEND_TELEMETRY_REQ form) with full permissions, but the app previously had no way to show it. This reuses the existing TelemetryScreen with an isSelf mode:

- Sends the self request instead of routing over the mesh, with a 5 second timeout since there is no sent confirmation for local requests
- Response matching works through a synthesized contact carrying the self public key
- Hides the routing mode button (meaningless for the local node) and skips the repeater battery snapshot cache
- Auto refresh works as it does for contacts

Entry points: a new Telemetry tile on the Settings screen (below Radio Stats, disabled when disconnected) and a thermostat icon on the Radio Stats app bar. Also shows humidity values with a percent sign, which applies to contact telemetry too.

No new localization keys, everything reuses existing strings.

## Testing

The self telemetry screen itself is hardware tested on Android (CI built APK) with a RAK4631 BLE companion running current MeshCore dev firmware with an SHTC3 sensor: battery voltage, MCU temperature, sensor temperature and humidity all displayed. The humidity percent formatting and the settings tile from the second commit are compile verified on all five platform targets in CI (analyze and format gates included) and running installed on a test device.

### DIFF
--- a/lib/screens/companion_radio_stats_screen.dart
+++ b/lib/screens/companion_radio_stats_screen.dart
@@ -1,6 +1,11 @@
+import 'dart:typed_data';
+
 import 'package:flutter/material.dart';
 import 'package:meshcore_open/connector/meshcore_connector.dart';
+import 'package:meshcore_open/connector/meshcore_protocol.dart';
 import 'package:meshcore_open/models/companion_radio_stats.dart';
+import 'package:meshcore_open/models/contact.dart';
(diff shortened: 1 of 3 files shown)


zjs81:
@codex review

PR #514, opened 2026-08-06
Show connected node hardware and firmware in device info

Hacuchino-hash:
## Summary

The DEVICE_INFO response from the firmware includes the board manufacturer name (bytes 20 to 59) and the firmware version string (bytes 60 to 79), but the connector skipped both fields. This parses them and shows two new rows, Hardware and Firmware, in the expandable Device Info card on the Settings screen, so you can see what hardware and firmware the connected node runs (the same way Meshtastic surfaces its hw_model, but for the local node using data MeshCore already sends).

Adds two localization keys (settings_infoHardware, settings_infoFirmware) with English fallback in the generated files, following the template arb.

## Testing

Compile verified on all five platform targets in CI, and the parsing follows the firmware frame layout (manufacturer at bytes 20 to 59, version string at 60 to 79, both already sent by the firmware today). Analyze and format gates pass in CI, and the build is running installed on a test device against a RAK4631 companion.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -271,6 +271,8 @@ class MeshCoreConnector extends ChangeNotifier {
   bool _batteryRequested = false;
   bool _awaitingSelfInfo = false;
   bool _hasReceivedDeviceInfo = false;
+  String? _manufacturerName;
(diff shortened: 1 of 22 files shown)


zjs81:
Looks good thank you

PR #515, opened 2026-08-06
Fix path trace test and restore CI checks

Hacuchino-hash:
## Summary

CI on dev is currently red: `flutter analyze --fatal-infos --fatal-warnings` fails because `MapTileCacheService` gained a required `appSettingsService` constructor parameter and `test/screens/path_trace_test.dart` still constructs it with no arguments. This wires the test's provider the same way `main.dart` does.

Since analyze fails before the formatting step ever runs, some format drift had also accumulated unnoticed. This includes a `dart format` pass over `lib` and `test` so both gates come back green.

## Testing

Validated on a fork run with the same gates this repo uses: full `flutter analyze --fatal-infos --fatal-warnings` and `dart format --set-exit-if-changed` both pass, and all five platform builds succeed.

### DIFF
--- a/lib/models/app_settings.dart
+++ b/lib/models/app_settings.dart
@@ -76,8 +76,7 @@ class Cyr2LatProfile {
 
 class AppSettings {
   static const Object _unset = Object();
-  static const String stadiaDemo =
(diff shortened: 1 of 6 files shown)


zjs81:
@codex review

PR #516, opened 2026-08-06
Warn when the node reports contact storage is full

Hacuchino-hash:
## Summary

The firmware pushes PUSH_CODE_CONTACTS_FULL (0x90) when the node's contact storage fills up, but the app ignored it, so on a busy mesh new nodes silently stop appearing with no explanation. This handles the push and shows a warning banner at the top of the contacts list. The flag clears on disconnect and automatically once a contact sync shows capacity has been freed.

One new localization key (contacts_storageFull) added through the template arb with English fallback in the generated files.

## Testing

Verified end to end on hardware. A RAK4631 companion was flashed with a local test build that calls onContactsFull() on a timer, so the firmware emits the real PUSH_CODE_CONTACTS_FULL frame, and the banner appeared on the contacts screen as expected. Normal operation with a healthy contact table shows no banner. Also compile verified on all five platform targets in CI with the analyze and format gates passing.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -271,6 +271,7 @@ class MeshCoreConnector extends ChangeNotifier {
   bool _batteryRequested = false;
   bool _awaitingSelfInfo = false;
   bool _hasReceivedDeviceInfo = false;
+  bool _contactsFull = false;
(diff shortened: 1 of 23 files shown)


zjs81:
@codex review

PR #517, opened 2026-08-06
Add high voltage LiPo battery chemistry option

Hacuchino-hash:
## Summary

Adds a LiPo HV (3.0-4.35V) battery chemistry option to both chemistry pickers (per device and per repeater). High voltage cells rest around 4.3V at full charge, for example the Seeed SenseCAP MeshTracker X1 ships one, and with the existing 4.2V ceiling the battery gauge pins at 100 percent for the whole top of the discharge curve.

One new localization key following the existing battery label pattern.

## Testing

Compile verified on all five platform targets in CI with the analyze and format gates passing, and running installed on a test device. The X1 hardware referenced reads 4.30V at rest when full, which the new range maps to 100 percent instead of overflow.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -343,6 +343,7 @@
   "appSettings_batteryNmc": "18650 NMC (3.0-4.2V)",
   "appSettings_batteryLifepo4": "LiFePO4 (2.6-3.65V)",
   "appSettings_batteryLipo": "LiPo (3.0-4.2V)",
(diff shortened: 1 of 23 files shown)


shortwavesurfer2009:
This will be somewhat helpful. I have a Heltec MeshPocket 10,000 mAh and at full charge it insists that it is at 4.48 volts. So while it will still cause the battery gauge to be pinned at 100%, it will be pinned at 100% for less time.

zjs81:
Looks good thank you

Issue #415, opened 2026-04-28
Idea: Image/File sideloading using IPFS


ericszimmermann:
Hello

I have just seen that the PR #223 is closed.
A grievance I had with this solution is, that it relied on a maintained asset server, so there was just a random thougt about:

**Using IPFS** as a sideload server for Images and/or Files? On Upload the files will get an Hash-Identifier (46 char long)
https://ipfs.tech/ https://docs.ipfs.tech/
There are also flutter plugins for ipfs.

Example file from a global gateway: https://dweb.link/ipfs/QmPvexzWa37f7HfvQ3mNuhGE9C9ALEHxC5Hc96PVjNgN55

What's the verdict?

446564:
I like this much better, at least it may be internet based but decentralized. And this way each device will be a node, correct?

ericszimmermann:
That depends on the Package:
- The first I looked at was [flutter_ipfs](https://pub.dev/packages/flutter_ipfs).
  But there were no details and the gitrepo seems to not be accessible anymore. So I would not recomend this package.
- The second was [ipfs_client_flutter](https://pub.dev/packages/ipfs_client_flutter).
  This is only a client, where you would have to run a node on your device or point to a public gateway.
- **The one I find most promising is [dart_ipfs](https://pub.dev/packages/dart_ipfs)**,
  which is actively developed and seems to have full p2p node.

benallfree:
Cool idea! I didn't know about IFPS, I just whipped up the cloudflfare solution as a demo. 

If we like IFPS, it probably wouldn't take much to dust off my PR and retrofit. It might still be a good idea to run an official trusted gateway and add a setting for users to specify their own gateway.

ericszimmermann:
I've tried to create a demo app using dart_ipfs, but haven't had any success so far.
Downloading already known file like the image from above works, but uploading and accessing the provided files do not yet.

ericszimmermann:
So, running full node on mobile didn't work.
You have to use a self hosted kubo ipfs node or a service like pinata.cloud or filebase.com (both have a free plan).
I wrote a Testimplementation for these: [IPFS-Photo-relay](https://github.com/ericszimmermann/IPFS-Photo-relay)
~~I would recoment to send "ipfs://\<cid\>" as text in meshcore app and try to download if this prefix is received.~~
It can also be used for files as for image which could be displayed directly.

ericszimmermann:
[...]

PR #518, opened 2026-08-07
Show images to detected URLs

ericszimmermann:
## Introduction
This is an implementation, which will show the Images which were send as url-link in messages.
This PR is the receiving end of the Discussion around sending images from #415 [Idea: Image/File sideloading using IPFS](https://github.com/zjs81/meshcore-open/issues/415), which was based on the closed PR #223 [feat: image sideloading](https://github.com/zjs81/meshcore-open/pull/223).

## Function
The function checks the messages for possible image urls.
Sending urls to pictures is not new and is already done regularly from multiple users from different clients.
It is only that with this PR they will be shown directly without you having to open the url in a browser.
The Implementation is provider agnostic and extensible. For now I recognize urls in the format:
- General Link:
http(s)://site.com/*** **.png .jpg .jpeg or .webp**

- **IPFS Gateway url:**
http(s)://gateway.com **/ipfs/** cid

- IPFS URI scheme (through https://ipfs.io/ipfs/cid):
**ipfs://** cid

- And since I got some links for from pyx.li and imgbb.com through the public channel!
for these provider the page url will be searches for the image paths:
https://pyx.li/?i=id
https://ibb.co/id

## Consideration
The Pros for sending Images through url is, that it is human readable. This means it is also readable by other clients and not some cryptic nonsense for others.
The Cons is, that it takes a lot of unnecessary chars.

## left out: Sending
What is not in this PR, is how to upload and send an image.
This is still in Discussion in #415 [Idea: Image/File sideloading using IPFS](https://github.com/zjs81/meshcore-open/issues/415) with my proposal for IPFS.
But tit has not to be IPFS, since we could be open to multiple image providers as long as they provide an url link directly to the image.
Nonetheless I have here at [IPFS-Photo-relay](https://github.com/ericszimmermann/IPFS-Photo-relay) a sample implementation for the IPFS pinning services filebase.com, pinata.cloud (free accounts available) and self hosted IPFS node.


ericszimmermann:
## UI
You can also tap onto the picture and it opens in a fullscreen view.
Here the preview in the normal message flow:
Screenshot_20260807_203137_com_meshcore_meshcore_open_MainActivity_50

shortwavesurfer2009:
Is there a way to disable this? Especially in public channels? Because I could see somebody potentially trying to use an image as a... exploit, like remote code execution, or perhaps just unsavory photos in the public channel, like... site.com/PornographicPhoto.png

ericszimmermann:
Yes I could implement a setting for this. I think I would try to implement it as channel setting and not globally, if that is acceptable.

shortwavesurfer2009:
> Yes I could implement a setting for this. I think I would try to implement it as channel setting and not globally, if that is acceptable.

How would implementing that as a channel-specific setting work for direct messages? And would it be automatic deny? Or automatic allow?

ericszimmermann:
I do not want to have to many settings.
And since meshcore is unmoderated I think it would be better as Opt-in.
With that you would have to enable for the persons and channels you feel comfortable with.
Opt-in is unfortunately not advertising that we have this option, especially if its in Channel setting instead of App settings. But I can live with that.

shortwavesurfer2009:
> Opt-in is unfortunately not advertising that we have this option

You could show like a generic image thumbnail in the chat and when tapped it could bring that setting up to turn on image downloading for that channel.

You know like the image you get on platforms when you forget to upload an image? Where you get like a gray square with a picture of mountains in a frame or whatever?

Edit: similar to this one. https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fcdn-icons-png.flaticon.com%2F512%2F9686%2F9686582.png&f=1&nofb=1&ipt=47a131ca3d34f1cf0f5bbb5de40ab994e11f607cb9590c60fabdb56b046383bb

ericszimmermann:
A thumbnail images was to big and to much clutter, I only added a small hint for possible image to be enabled in Settings.

zjs81:
Is this ready to be merged?

zjs81:
@codex review

ericszimmermann:
From my side yes.
I see you added a codex review, so there may be some finds.

zjs81:
> From my side yes. I see you added a codex review, so there may be some finds.

Thank you! Could you fix those findings and I'll take another look.

ericszimmermann:
Sorry, after the third try I got it to build android again. He still throws a warning that you should update grade to at least 9.1.0, because support for 8.14.0 will soon be dropped, but it builds.

The points above should now be resolved. At least on my test run I had no issues.

zjs81:
Thanks just merged

Issue #257, opened 2026-03-04
Feature : An option to hide the GIF button


neopiccolorat:
Maybe I'm alone with this one, but I'm not a fan this GIF button.
I don't think it deserves this prime place on the UI. Just my opinion of course :)

ericszimmermann:
Well if the option #223 (image sideloading) is comming we could make the button an "Action button" under which we can have a pop-up for gifs, Image-sideloading and others.

ericszimmermann:
maybe also add the emoji picker from widget/emoji_picker?

neopiccolorat:
Yes, I personally would prefer that, the emoji picker (that that I "like" emojis) but it's more in line with every other keyboard/messaging app

ericszimmermann:
I opened a new Pull-request #346, since the old one got quite a commit history.

PR #522, opened 2026-08-13
clean up input bar

ericszimmermann:
I made a small change to unclutter the InputBar at the bottom of chat_screen and channel_chat_screen.
So that we have more space to actually type something.

The send_Gif and send_Image Buttons (and maybe further send Actions like emoji-picker and location) are now inside a Pop-Up menu.
(Discussion in #257)

For better overview, I renamed the _buildMessageComposer from channel_chat_screen
to _buildInputBar like in the chat_screen, so that they are similar.
I also looked that they have now the same structure with safearea and padding.


ericszimmermann:
This Pull-Request is based on zjs81/dev 724696cd3c5f12be15d5dd3a5f204e2cd2585f5d,
which tested will already not build for android with error:
```An exception occurred applying plugin request [id: 'dev.flutter.flutter-gradle-plugin']
> Failed to apply plugin 'dev.flutter.flutter-gradle-plugin'.
   > Error: Your project's Gradle version (8.12.0) is lower than Flutter's minimum supported version of 8.14.0. Please upgrade your Gradle version. 
     Alternatively, use the flag "--android-skip-build-dependency-validation" to bypass this check.
     Potential fix: Your project's gradle version is typically defined in the gradle wrapper file. By default, this can be found at /home/runner/work/meshcore-open/meshcore-open/android/gradle/wrapper/gradle-wrapper.properties. 
     For more information, see https://docs.gradle.org/current/userguide/gradle_wrapper.html.```

zjs81:
This looks good code wise. Could you send a screenshot of what this change looks like?

ericszimmermann:
Here is a screenshot of the pop-up:
image
the right side with the send button is unchanged.

zjs81:
Looks great thank you

Issue #521, opened 2026-08-12
[Feature request] Support MeshCore One (iOS app) Reply and Reactions


venamartin:
Feature request: read MeshCore One replies and reactions

MeshCore One (iOS) sends replies and reactions as plain, human-readable text.

Reply format

```
@[{target name}]
>{first ~10 chars of the quoted message}..
{reply body}
```

Example on the wire:

```
@[GWQ∆🍓]
>Heading to..
Nice!
```

A client that parses it can render a proper quote bubble. A client that
doesn't still shows a readable reply.

## Reaction format

```
{emoji}@[{target sender}]     <- channel form; DMs omit "@[...]"
{hash}
```

Example: `👍@[GWQ∆🍓]` + newline + `kryv4zmp`

**Hash:** first 5 bytes of `SHA-256(UTF-8 text + uint32-LE sender timestamp)`,
https://github.com/Avi0n/MeshCoreOne/blob/main/docs/Reactions.md

Bubbles can still be generated and reactions can find the message and place the reaction emoji on the message. This method really makes it more readable on ALL meshcore apps and hardware.

sethoscope:
I have reactions working. I'll submit a PR after I write a few more tests.

Related: #452

Avi0n:
FYI I'll be switching MC1's reaction format slightly in the next version to hopefully make it more clear to other users.

New format:
`@[{targetSenderName}]{emoji}\n{hash}`
Example:
`@[AlphaNode]👍\nb45pc4ek`

The goal is to reduce confusion by placing the emoji after the `@` since that's what is normally done/expected. I think it's possible that people were misinterpreting the hash as the thing they were being replied to with instead of the emoji.

venamartin:
@Avi0n I'll follow with MeshTrax. Support both on receiving and @ first on transmit. 

@Avi0n 👍

PR #537, opened 2026-08-30
Handle MeshCoreOne's emoji reactions

sethoscope:
MeshCoreOne is a companion app that supports emoji reactions, using a different message format than ours.

This commit adds support for parsing and displaying reactions received in the MeshCoreOne format.

Their format is described here:
 https://github.com/Avi0n/MeshCoreOne/blob/main/docs/Reactions.md

Here's an example:

👍@[entropy]
b3je6qf7

Major differences compared to MeshCore-Open's reaction messages:
- includes the emoji, rather than an index code
- includes the name of who sent the message it refers to
- the hash does not depend on the sender name

See #521.

### DIFF
--- a/lib/connector/meshcore_connector.dart
+++ b/lib/connector/meshcore_connector.dart
@@ -3563,27 +3563,31 @@ class MeshCoreConnector extends ChangeNotifier {
 
       // Process reaction locally to update the UI immediately
       appLogger.info('Adding sent channel reaction, id: $reactionIdentifier');
-      _processReaction(messages, reactionInfo);
-      await _channelMessageStore.saveChannelMessages(channel.index, messages);
+      final reactionApplied = _processReaction(messages, reactionInfo);
+      if (reactionApplied) {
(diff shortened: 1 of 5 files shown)


Avi0n:
FYI I'll be switching MC1's reaction format slightly in the next version to hopefully make it more clear to other users.

New format:
`@[{targetSenderName}]{emoji}\n{hash}`
Example:
`@[AlphaNode]👍\nb45pc4ek`

The goal is to reduce confusion by placing the emoji after the `@` since that's what is normally done/expected. I think it's possible that people were misinterpreting the hash as the thing they were being replied to with instead of the emoji.

zjs81:
@codex review

sethoscope:
@zjs81 All set.

sethoscope:
This PR's handling of MeshCore-One's updated reaction format has now been tested in the wild, as they released their update recently. Might as well support both for a while, I think.

zjs81:
Thank you

PR #541, opened 2026-09-03
Adding preset for Philly per latest blog post

kyleyankan:
Adding a USA Philly area preset as per the blog post here: https://phillymesh.net/2026/09/02/fcc-regulations/

902.250mhz, BW500, SF11.

### DIFF
--- a/lib/models/radio_settings.dart
+++ b/lib/models/radio_settings.dart
@@ -482,6 +482,16 @@ class RadioSettings {
       ),
     ),
     (
+      'USA Philly',
+      RadioSettings(
+        frequencyMHz: 902.250,
+        bandwidth: LoRaBandwidth.bw500,
+        spreadingFactor: LoRaSpreadingFactor.sf11,
+        codingRate: LoRaCodingRate.cr4_5,
+        txPowerDbm: 20,
+      ),
+    ),
+    (
       'USA/Canada',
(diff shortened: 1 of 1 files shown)


RightsizedAlpaca8594355:
And Coding Rate = 4/5! Thank you!

shortwavesurfer2009:
> And Coding Rate = 4/5! Thank you!

The coding rate is the least important metric, as all it does is add error correction, and any node can understand any node coding rate. So, if you use 4/8 for the preset, you get the longest distance possible, with the sacrifice of speed, and if the user is close enough, then they can choose to go 4/5

RightsizedAlpaca8594355:
Agreed fully on all accounts. We want nodes to start at (aka default to) CR 4/5 to minimize unnecessary airtime. And then for fringe suburb connections we can individually ratchet those specific nodes up towards 4/8. But a majority of nodes will work in the urban mesh at default 4/5.

Thanks!
-Alpaca

-------- Original Message --------

dustyheatsink:
These settings are working well for our mesh, and we have a lot of users switching from meshtastic to meshcore because of it. currently they are asking us for app recommendations, and then we are walking them through the settings based on their choice. having our region be an available preset would greatly simplify the process, thanks for the consideration!

PR #544, opened 2026-09-07
Updated Swedish language file

DomeDan:
I've updated the swedish language file with help from other swedish members of the Sweden-Mesh discord server.
Many of the previous translations were directly translated word by word and were very confusing to read in the app.


zjs81:
Thanks!

PR #545, opened 2026-09-08
Add PhillyMesh 500 kHz radio preset

zjs81:
## Summary

Adds a Philadelphia/PhillyMesh 500 kHz radio preset to the existing regional preset list.

Preset values supplied by the PhillyMesh community:
- Frequency: 902.250 MHz
- Bandwidth: 500 kHz
- Spreading factor: SF11
- Coding rate: 4/5
- TX power: 20 dBm (consistent with the existing US presets)

The preset is labeled `USA Philadelphia (PhillyMesh 500 kHz)` so Philadelphia-area users can select the correct configuration directly from Radio Settings.

No BLE protocol changes.


shortwavesurfer2009:
I must say that I'm curious as to why you made your own pull request instead of going with https://github.com/zjs81/meshcore-open/pull/541

zjs81:
> I must say that I'm curious as to why you made your own pull request instead of going with #541

Didn't see it. I just merged that one in.

Issue #464, opened 2026-05-25
Sync clock sets incorrect time


todd-herbert:
I have recently had issues with setting repeaters' clocks. After sync, the remote device's clock gives a time ~15 years in the future.

I *think* this issue appeared in alpha9, although I'm not 100% sure.

Please let me know if more information is needed.

hdhog:
i have similar problem but jump to one year. official app sync time correctly.

todd-herbert:
> official app sync time correctly.

If I remember correctly, when I tested with Liam Cottle's app, the sync time button worked, but issuing the clock sync command via the cli did not.

hdhog:
my be this commit in firmware affected https://github.com/meshcore-dev/MeshCore/commit/4a869163 and now needed use `time <epoch_seconds>` instead `clock sync`  similar as meshcore_cli.py

PR #548, opened 2026-09-09
fix: use explicit time <epoch> for repeater clock sync

hdhog:
Companion firmware (MeshCore commit 4a869163, 2025-12-30) replaces the CLI frame timestamp with the companion node's RTC for TXT_TYPE_CLI_DATA, so "clock sync" sets the repeater clock to the node RTC. Nodes with a wrong RTC push repeater clocks years into the future; firmware cannot set the clock backwards without clkreboot.

- RepeaterCommandService: translate "clock sync" to "time <phone epoch>" before sending, matching official meshcore-cli behavior
- repeater login dialog: send "time <epoch>" for auto clock sync after login (this path bypasses RepeaterCommandService)
- add unit tests for the command translation

Ref: https://github.com/zjs81/meshcore-open/issues/464


zjs81:
Looks good thank you

PR #549, opened 2026-09-10
max power on USA presets

### LINKED ISSUE #466: raise tx power for US/Canada
The US and canada can do +30dbm on the 902-928 (915MHz (33cm) industrial, scientific, and medical) band. I am fully aware that most radios only go as high as +22dbm with more recent radios and the station G2 being the exceptions.

Therefore the US Arizona and US/Canada preset can be bumped to +22dbm instead of the +20 it is at now.

I am not suggesting raising the power on the off-grid one that switches you to 918 MHz because it can also be used in other countries that may have different power limits.

That change increases the tx power by 58.5mW 0.1W to 0.1585W

shortwavesurfer2009:
this changes the max power on all USA presets to 22 instead of 20. in the US we are allowed a max of 30 so this is within limits

Edit: closes https://github.com/zjs81/meshcore-open/issues/466


zjs81:
thanks

PR #550, opened 2026-09-10
Add compliant preset

### LINKED ISSUE #542: [Feature] change US/Canada to be in compliance
Turns out that the FCC requires a minimum bandwidth of 500 kilohertz and so the current setting defaulting to 62.5 kilohertz makes this out of spec.

https://github.com/meshcore-dev/MeshCore/issues/945#issue-3516426536

I recommend having two presets, one called 
"US / Canada (recommended)
Freq 910.100 MHz
BW 500KHz
SF 10
CR 8

And the other would be...
"US / Canada (legacy)"
Freq 910.525
BW 62.5
SF 7
CR 8

This would put the compliant setting as recommended, but still leave the option for users to choose the other preset for systems during the transition period. The legacy preset could then be removed in a couple of versions in order to not have it as an option. I am not aware of any FCC fines resulting from running out of compliance but it seems like it would be best to start making a transition to compliant settings.

shortwavesurfer2009:
This PR adds an FCC compliant preset as "recommended" and keeps the current preset as "legacy". This would resolve https://github.com/zjs81/meshcore-open/issues/542


LitBomb:
No one in any official FCC capacity has publicly said any LoRa system is non-compliant to any FCC rules.  All these non-compliant claims are unsubstantiated.

CullenShane:
@shortwavesurfer2009 Without coordination from the main project, this can't be called "Recommended" it must be called something else until it is actually the recommended settings.  I suggest "Compliant" or "500Khz" or something so established meshes or the main project can respond without causing fragmentation.

shortwavesurfer2009:
> @shortwavesurfer2009 Without coordination from the main project, this can't be called "Recommended" it must be called something else until it is actually the recommended settings. I suggest "Compliant" or "500Khz" or something so established meshes or the main project can respond without causing fragmentation.

You make a good point. I will change that from "recommended" to "FCC compliant" and keep the old one as "legacy" as it is set with this PR. Give me a few minutes to get those changes put in.

shortwavesurfer2009:
I did end up going with "compliant" instead of "FCC compliant" because I forgot that the Canadian agency is not the FCC even though they follow the FCC guidelines.

shortwavesurfer2009:
This change also does not overlap with any of the normal meshtastic modem settings.

https://meshtastic.org/docs/overview/radio-settings/

PR #554, opened 2026-09-12
Add contact import from QR code

### LINKED ISSUE #529: Add Contacts from a QR Code from a Picture
is this Possible to add Contacts from a QR Code Picture this was nice ..

M0Rf30:
Fixes #529

## Problem

Contacts can only be imported by pasting `meshcore://<hex advert packet>` from the clipboard (`contacts_screen.dart` `_contactImport`). Contact QR codes and links produced by the official MeshCore app use a different format and cannot be used at all:

```
meshcore://contact/add?name=Example+Contact&public_key=<64 hex>&type=1
```

(format documented in the firmware repo, `docs/qr_codes.md`)

## Solution

* `lib/utils/contact_qr.dart` — pure-Dart `parseContactQr()` returning a `ScannedContact`. Accepts **both** the official contact-card link and the existing `meshcore://<hex>` advert form; returns `null` for anything else (including `meshcore://channel/add?...`). The public key must be exactly 64 hex chars; `type` falls back to `1` (Companion) when missing or out of the documented 1..4 range. Advert blobs are bounded to 98..171 bytes — the protocol minimum, and `maxFrameSize` minus the command byte.
* `lib/utils/contact_qr_frame.dart` — `buildContactQrFrame()` selects the device frame: advert blobs replay through `CMD_IMPORT_CONTACT`; contact cards are written with `CMD_ADD_UPDATE_CONTACT` using the `0xFF` unknown-path sentinel (a card carries no route, so messages flood until an advert supplies one), preserving the flags and name of an existing contact and truncating the name on a UTF-8 character boundary.
* `lib/screens/contact_qr_scanner_screen.dart` — new screen modelled on the existing `CommunityQrScannerScreen`: same app bar, themed corner overlay and snackbar conventions, reusing `QrScannerWidget`. Besides the live camera it has a gallery action (`image_picker` + `MobileScannerController.analyzeImage`), which is the literal request in #529 ("from a Picture"). Writes are sent with `waitForGenericAck: true` and followed by `refreshContacts()` so the imported contact is visible on return.
* Entry points: one item in the contacts overflow menu and one in the ⊕ add-contact sheet. The clipboard path is untouched.

No new dependencies — `mobile_scanner`, `image_picker` and `qr_flutter` are already in `pubspec.yaml`. No manifest changes: `CAMERA` is already declared for the community scanner, and `image_picker` supplies its own storage permission via manifest merge.

## Localization

Five new keys, translated in **English and Italian**; the other 37 locales fall back to English until translated, and `untranslated.json` tracks them accordingly.

## Tests

* `test/utils/contact_qr_test.dart` (20 cases): official card with `+`/percent-encoded name, uppercase key, explicit and defaulted `type`, 63-char / non-hex / sign-prefixed keys rejected, legacy hex form, 97-byte and 172-byte boundaries rejected with 98 and 171 accepted, `channel/add`, foreign scheme, empty and whitespace input.
* `test/utils/contact_qr_frame_test.dart` (5 cases): `CMD_IMPORT_CONTACT` opcode and untouched advert payload; card frames asserted byte-by-byte (public key, type, flags, `0xFF` path sentinel, name), flags preserved on re-scan, name fallback, and UTF-8 boundary truncation.

## Verification

* `dart format .` clean
* `flutter analyze` — No issues found
* `flutter test test/utils/` — 39/39 pass
* full `flutter test` — only the two pre-existing `message_url_image_helper_test.dart` pyx failures, which also fail on an unmodified `dev` checkout

Generated l10n output is included for the new keys only; no unrelated regeneration drift.

## Notes

* `analyzeImage` is unsupported on the iOS Simulator; that path shows the "no QR code found" snackbar instead of throwing.
* Gallery decoding distinguishes three outcomes: no barcode in the image, a barcode that is not a MeshCore contact, and a picker/permission failure.

---

AI disclosure per CONTRIBUTING: the patch was written with agent assistance and reviewed by me; scope was kept to the files listed above.


M0Rf30:
Thanks — all six findings addressed in the amended commit.

**Critical, `contact_qr_scanner_screen.dart:208` — ACK not awaited.** Correct: `expectsGenericAck` only queues the frame. Switched to `waitForGenericAck: true`, matching `MeshCoreConnector.importDiscoveredContact` (`meshcore_connector.dart:3871`). Firmware errors now propagate into the existing failure snackbar instead of the screen reporting success and popping early.

**Moderate, `:214` — cached contacts not refreshed.** Added `await connector.refreshContactsSinceLastmod()` after the ACK, so the imported card is present in `connector.contacts` before the pop and `ContactsScreen` renders it immediately.

**Moderate, `:199` — flags clobbered on re-scan.** The card path now looks the contact up by `publicKeyHex` in `connector.contacts` and passes `flags: existing?.flags ?? 0`; an empty `name` in the QR also falls back to the existing name rather than blanking it.

**Moderate, `:146` — only the first gallery barcode inspected.** Now iterates `capture.barcodes` and takes the first value that `parseContactQr` accepts, so an unrelated barcode in the photo no longer masks a valid contact QR.

**Moderate, `contact_qr.dart:27` — advert minimum too low.** Raised to the protocol's 98 bytes (`_minAdvertPacketBytes`), matching `buildImportContactFrame`. Boundary test added: 97 bytes rejected, 98 accepted.

**Moderate, `contact_qr.dart:79` — `int.tryParse` accepts signs.** Real bug, good catch: `+1` parsed as 1 and `-1` would have thrown on the `Uint8List` assignment. Replaced with an explicit `_hexDigit` code-unit check over `[0-9a-fA-F]`; regression test `signed public key digits are rejected` covers a `'+1' * 32` key.

`flutter analyze` clean, `test/utils/contact_qr_test.dart` 16/16 pass.

M0Rf30:
Second round addressed.

**`contact_qr_scanner_screen.dart:213` — `path_len` 0 means zero-hop, not unknown.** Correct and the most consequential one: a contact card carries no route, so a direct-only contact behind a repeater would have been unreachable until an advert arrived. Now sends `0xFF`, the sentinel `Contact.fromFrame` maps to `hopCount == -1` (`contact.dart:171`), so initial messages flood. Comment added at the callsite. This also moots the `_encodePathLenForCurrentMode` point from the previous round: that helper encodes the hash-width mode in bits 6-7 for a *real* path, which does not apply to the unknown-path sentinel.

**`:222` — `refreshContactsSinceLastmod()` can skip the record.** Agreed, the high-water mark is not a safe filter when the frame is stamped with the phone clock. Switched to `connector.refreshContacts()` (full list, `preserveExisting: true`).

**`contact_qr.dart:51` — no upper bound.** Bounded to 171 bytes (`_maxAdvertPacketBytes`), i.e. `maxFrameSize` minus the command byte `buildImportContactFrame` prepends, so an oversized QR is rejected at parse time instead of failing at transport. Boundary tests both ways: 171 accepted, 172 rejected.

**`:169` — 'no QR code found' shown for a decoded but unsupported payload.** Now tracks `decodedAnyBarcode` while scanning the capture: a decoded non-MeshCore QR gets 'Invalid contact data', and the new message is reserved for images with no barcode at all.

**`:133` — picker failure misreported as invalid contact data.** Added `contacts_qrGalleryFailed` ('Could not open the gallery.') with a dedicated `_showGalleryError()`, so permission/platform failures are no longer diagnosed as bad QR content.

**`untranslated.json:139` — Swedish inconsistency.** Root cause was the opposite of a tracking-file mistake: `app_sv.arb:493` has had `"reaction_report": "Emoji-reaktioner"` since #544, but the committed `app_localizations_sv.dart` was never regenerated and still returned the English fallback, so the generator legitimately drops the key from the untranslated set. Rather than restore a stale entry I applied the one-line generator output (`reaction_report => 'Emoji-reaktioner'`), which makes the artifacts consistent without pulling in the ~850 lines of unrelated Swedish regeneration this PR should not carry.

**`contact_qr_test.dart:25` — "String has no `*` operator".** This one is a false positive: `String.operator *` exists in `dart:core` (`'ab' * 32`). The file compiles and the suite passes — 18/18 after the two new boundary cases.

Italian added for all five new keys (`app_it.arb`), plus the QR entry is now also in the ⊕ add-contact sheet, not just the overflow menu.

`flutter analyze` clean.

M0Rf30:
Third round addressed.

**`:130` — gallery picker races the camera.** Real: `_isProcessing` was only claimed *after* `pickImage` returned, so a camera detection could enter `_handleScannedData` while the picker sheet was open and produce a second import or a misleading late error. The flag is now set before awaiting the picker and cleared on both cancellation and failure, so exactly one import can be in flight.

**`:212` — import branch untested.** Extracted the frame selection into `lib/utils/contact_qr_frame.dart`:

```dart
Uint8List buildContactQrFrame(ScannedContact scanned, {Contact? existing})
```

The screen is now a thin caller, and `test/utils/contact_qr_frame_test.dart` asserts the wire layout directly rather than through a fake connector: advert blobs emit `cmdImportContact` followed by the untouched packet; cards emit `cmdAddUpdateContact` with the public key at offset 1..32, type at 33, flags at 34 and `0xFF` at 35 (the regression guard for the zero-hop/unknown-path bug from the previous round), plus the flags-preserved and name-fallback cases. 38 tests pass.

I did not add a widget test for the ACK-failure and refresh paths: that needs a `MeshCoreConnector` fake, and the class is a concrete `ChangeNotifier` wired to BLE/USB transport with no interface to implement, so the test would end up asserting against a mock of my own construction rather than real behaviour. `sendFrame(..., waitForGenericAck: true)` and `refreshContacts()` are both existing, already-exercised connector APIs — the value was in verifying the frame this screen hands them, which the new test does. Happy to add the widget test if you would rather see it, or if a connector seam is something you want introduced separately.

`flutter analyze` clean.

M0Rf30:
**`contact_qr_frame.dart:23` — UTF-8 name truncation.** Confirmed against `BufferWriter.writeCString` (`meshcore_protocol.dart:133-136`): it copies encoded bytes up to `maxLength - 1` with no boundary check, so a name crossing 31 bytes mid-rune is stored malformed and comes back as replacement characters.

Fixed in `buildContactQrFrame` by dropping whole characters before the frame is built (same rune-walking approach as `Utf8LengthLimitingTextInputFormatter._truncateToMaxBytes`, which the app already uses for name input). Test added: a 16-character name of three-byte runes (48 bytes) yields exactly 30 bytes / 10 intact characters, and `utf8.decode` round-trips it.

Scoped to this call site rather than `writeCString` itself: that writer is shared by channel names and every other contact write, so hardening it belongs in its own change with its own tests rather than riding along here. Happy to do that separately if you want it at the writer level.

`flutter analyze` clean, frame tests 5/5.

M0Rf30:
**Swedish drift — reverted as asked.** `app_localizations_sv.dart` is back to `reaction_report => 'Emoji Reactions'` and `untranslated.json` lists the key for `sv` again. Both files now contain additions for the five new QR keys and nothing else, so the PR carries no unrelated localization output.

For the record the underlying inconsistency is upstream, not introduced here: `app_sv.arb:493` has `"reaction_report": "Emoji-reaktioner"`, so a clean `flutter gen-l10n` on `dev` produces exactly the change I just reverted. Whoever regenerates next will reintroduce it. Worth a separate housekeeping commit.

**PR description — reconciled.** The 'only the English ARB is populated' line was stale; Italian was added later in the branch. Description rewritten to describe what actually ships, including the frame-builder extraction, the byte bounds and the UTF-8 truncation.

**`String` repetition operator — not a defect.** This is the third report of it, so concretely: `String` implements `operator *(int times)` in `dart:core` (`String.operator *`, present since Dart 1). `'ab' * 32` is valid and idiomatic. Both test files compile and run — `flutter test test/utils/` reports **39/39 passing**, which is not reachable if a file in that directory failed to compile. `flutter analyze` is also clean. I am not going to rewrite the fixtures as `List.filled(...).join()` to satisfy a claim the toolchain contradicts.

M0Rf30:
**`:226` — discovery-only nodes excluded from the lookup. Fixed.** Switched to `connector.allContactsUnfiltered`, which is `[..._contacts, ..._discoveredContacts]`, so a saved contact still wins (its flags matter) and a discovery-only record supplies the name when the card omits one. Comment added at the callsite.

**`contact_qr.dart:48` — opaque `meshcore:contact/add?…` also matches. Intentional, keeping it.** The normalization collapses `uri.host` and `uri.path` precisely so the parser does not depend on how a given input happens to split across the authority boundary. `Uri.parse` is not uniform here: whether `contact` lands in `host` or in `path` varies with the exact spelling a QR encoder emits, and QR payloads arrive from third-party apps, not from this codebase. Accepting a scheme-relative spelling of a well-formed contact card costs nothing — the payload still has to carry a valid 64-hex `public_key` or the parse fails — while tightening it risks rejecting legitimate cards produced by an encoder that normalizes differently. `channel/add` and every other path are still rejected, which is the distinction that matters.

**`test/utils/contact_qr_test.dart:25` — `String` repetition. Fourth report, still incorrect.** `String` implements `operator *(int times)` in `dart:core`; `'ab' * 32` is valid Dart. Verified again on this push: `flutter analyze` reports no issues and `flutter test test/utils/` reports **39/39 passing**, which cannot happen if a file in that directory fails to compile. I will not rewrite working fixtures to `List.filled(...).join()` on this basis.

M0Rf30:
Fifth report of the same claim, so here is the machine output rather than another assertion.

```
$ cat /tmp/s.dart
void main() {
  print('ab' * 3);
}

$ dart run /tmp/s.dart
ababab

$ dart --version
Dart SDK version: 3.13.3 (stable) (Tue Sep 1 01:07:17 2026 -0700) on "linux_x64"
```

`String.operator *(int times)` is part of `dart:core` and has been since Dart 1: <https://api.dart.dev/stable/dart-core/String/operator_multiply.html>

And the file the comment says cannot compile:

```
$ flutter test test/utils/contact_qr_frame_test.dart
00:00 +4: buildContactQrFrame contact card drops whole characters when the name exceeds 31 bytes
00:00 +5: All tests passed!
```

Both test files compile, run and pass; `flutter analyze` reports no issues across the whole package. Closing this thread out — no change will be made for it.

The two substantive findings from this round were handled in the previous push (discovery-only contact lookup fixed; the URI normalization deliberately kept lenient, with reasoning above). I believe the PR is complete and would appreciate a maintainer look whenever convenient.

zjs81:
@codex review

zjs81:
Can you please review the codex findings then I'll merge it in?

M0Rf30:
Reviewed the Codex round — one finding, it is correct, fixed and pushed (`d0583f2`). Also rebased on current `dev`, so the branch merges cleanly again.

**`contact_qr_frame.dart:21` (P2) — rescanning replaced a known route with flood.** Right, and it is the mirror image of the earlier `path_len` bug: `0xFF` is the correct value for a contact the device has never seen, but it was also being written for a contact that already has a usable path, so rescanning a card — or scanning one already discovered with a route — downgraded it to flood until the next advert arrived.

`buildContactQrFrame` now reuses the existing route when there is one:

```dart
({Uint8List path, int pathLen}) _routeOf(Contact? existing, int width)
```

It keeps `existing.path` and encodes `pathLen` the same way `MeshCoreConnector._encodePathLenForCurrentMode` does — hop count in bits 0-5, `pathHashByteWidth - 1` in bits 6-7 — and falls back to `0xFF` with an empty path only when there is genuinely nothing to preserve: no existing record, a flood-only contact (`pathLength < 0`), or a stored path that does not match the current path-hash mode (`path.length != hops * width`), which is the same validity rule the connector applies before sending. The scanner passes `connector.pathHashByteWidth` so a device configured for 2-4 byte hashes gets the right encoding, which also settles the earlier `_encodePathLenForCurrentMode` comment now that a real path can be written on this path.

Four tests added in `contact_qr_frame_test.dart`: two-hop route preserved in 1-byte mode, `0x42` encoding plus 4 path bytes in 2-byte mode, sentinel for a flood-only contact, sentinel when the stored path does not fit the mode. The existing "unknown-path sentinel" case still covers a brand-new contact.

Verification on the rebased branch: `flutter analyze` — No issues found; `flutter test` — 597 pass, with only the two pre-existing `message_url_image_helper_test.dart` pyx failures that also fail on an unmodified `dev`; `dart format` clean. The rebase conflict was `untranslated.json` only (both sides append keys); resolved additively, and `sv` follows `dev`, which has since regenerated its Swedish output, so this branch still carries no unrelated localization drift.

M0Rf30:
The red `Flutter and Dart` check on this PR is not from this branch: **597 tests pass, 2 fail**, and the two failures are `MessageUrlImageHelper parses short pyx ids` and `parses direct pyx image urls`.

Those tests fetch `https://pyx.li/?i=<id>` over the network and scrape its ``. Both uploads have expired — the page now returns a placeholder ("This page is currently in zen mode.") with no `` element, and `https://pyx.li/i/RzmdkTsE.jpg` is a 404 — so `parse` returns null. They fail identically on a clean `dev` checkout at 92e1d15, before any of my changes, and every open PR inherits the red check because the workflow only runs on `pull_request`.

Opened #556 against `dev` to remove them, with the evidence. Once that merges, re-running the checks here should be green; nothing on this branch needs to change for it.

PR #555, opened 2026-09-12
Show and copy the full node public key

M0Rf30:
## Problem

The device info card renders the node's own public key truncated to its first 16 hex characters:

```dart
// lib/screens/settings_screen.dart:305-306
value: '${pubKeyToHex(connector.selfPublicKey!).substring(0, 16)}...',
```

There is no reveal, no selection and no copy affordance, and the key is not shown in full anywhere else in the app. To actually obtain it a user has to either export their own advert to the clipboard (Contacts → Copy Advert to Clipboard) and manually slice the first 32 bytes out of the hex blob, or attach a serial terminal and run `get public.key`.

That matters whenever the full key is needed out-of-band: sharing an identity with someone not on the mesh yet, filling in a repeater ACL (`setperm <pubkey> …`), or pasting into the contact-card links the official app produces.

## Solution

Show the full 64-character key and copy it to the clipboard on tap, with a confirmation snackbar. `_infoRow` already accepts an `onTap` and wraps its content in an `InkWell`, so this is a value change plus a small helper — no layout changes for other callers.

The truncation was not protecting anything: the on-air identity hash is a **single byte** of this same key (`Identity::copyHashTo`, `PATH_HASH_SIZE == 1`), contact public keys are already displayed in full in the chat details sheet (`chat_screen.dart:872`), and the value was rendered unredacted — just clipped.

## Changes

* `lib/screens/settings_screen.dart` — full key as the row value, `onTap` copies it, new `_copyPublicKey` helper using `Clipboard.setData`.
* `lib/l10n/app_en.arb`, `lib/l10n/app_it.arb` — one new key, `settings_publicKeyCopied` (EN + IT), plus regenerated accessors.

One file of logic; no new dependencies, no protocol changes.

## Verification

* `dart format .` clean
* `flutter analyze` — No issues found

Not verified on a device: `selfPublicKey` is only populated once a node is connected. The 64-character mono value wraps to two lines inside the card; if you would prefer an explicit copy icon or a `SelectableText` instead of a tappable row, say so and I will adjust.

---

AI disclosure per CONTRIBUTING: written with agent assistance and reviewed by me; scope kept to the files listed above.


M0Rf30:
**Null assertion during disconnect — fixed.** The tap callback re-read `connector.selfPublicKey!`, which is cleared on disconnect, so a tap racing a disconnect (or a rebuild with the row still on screen) would throw on the bang operator.

The row is now built by a small `_publicKeyRow` helper that takes the already-rendered hex string and closes over it, so the copy uses the value the user is looking at and never touches the connector again:

```dart
onTap: () => _copyPublicKey(context, publicKeyHex),
```

The outer `if (connector.selfPublicKey != null)` guard still gates rendering, so there is no path left that dereferences a cleared key.

`flutter analyze` clean.

zjs81:
@codex review

zjs81:
thanks

PR #557, opened 2026-09-13
Close USB serial port off the UI isolate

M0Rf30:
## Problem

Connecting over USB to a device that never drains its serial input freezes the whole app for ~30s.

`close(2)` on a tty whose output the peer has not drained blocks in the kernel for the line discipline's `closing_wait` (30s default). `FlSerial.closePort()` is a synchronous FFI call, so `UsbSerialService.disconnect()` blocks the Dart UI isolate for that entire window.

Path: `connectUsb` writes APP_START -> 2x3s SELF_INFO timeout -> throws -> `disconnect(manual: false)` -> `closePort()` -> UI frozen ~30s. Even the 22 APP_START bytes are enough to wedge the close; the buffer does not need to be full.

Reproduced against a Seeed Wio Tracker L1 (`2886:1667`) flashed with `WioTrackerL1_companion_radio_ble`. That build has no `ENABLE_USB_INTERFACE`, so `examples/companion_radio/main.cpp` never wires `ArduinoSerialInterface` to `Serial` and nothing reads USB CDC: raw pyserial gets 0 bytes in 4s, a 1-byte write times out, and a raw `O_NONBLOCK` write loop queues 8192 B into the kernel with 0 bytes drained by the device. Wrong firmware is a user error, but the app should report a failed connect, not hang.

## Change

- `disconnect()` calls a new `_closePortOffUiIsolate()` which runs `bindings.fl_close(flh)` through `Isolate.run` (`fl_close` only touches process-global native state), then replicates the Dart-side cleanup `closePort()` performs: close `onSerialData`, `calloc.free` the two 8 KB I/O buffers, reset `flh`.
- Warn in the app debug log when the native close exceeds 1s, so "device is not draining its serial input" becomes visible instead of a silent freeze.
- `dispose()` intentionally stays synchronous: hot restart needs the native `SerialThread` killed before the isolate is torn down.
- Adds `ffi` to `pubspec.yaml` (already transitive) for `calloc.free`.

## Verification

Throwaway probe against the real wedged `/dev/ttyACM0` (8192 B queued and undrained), 100 ms periodic timer on the main isolate:

| close mode | duration | main-isolate timer ticks during close |
| --- | --- | --- |
| inline (current behaviour) | 30476 ms | **0** |
| `Isolate.run` (this PR) | 30302 ms | **302** |

The kernel still waits out `closing_wait`; the UI no longer does.

`flutter analyze lib/services/usb_serial_service_native.dart` -> no issues.

No in-app run: `flutter build linux` fails in my environment before Dart compilation, in flserial's vendored `tinycthread.c` vs glibc >= 2.42 (`once_flag` typedef redefinition / conflicting `pthread_once`). Confirmed pre-existing on an unmodified tree.

## Follow-ups (not in this PR)

In the MeshEnvy flserial fork:
- `src/flserial.cpp:155-157` skips `setTimeout` on Linux/macOS, so `Timeout()` is all zeros: the `SerialThread` busy-spins on `read()` for the whole session, and `write()` silently drops frames (`unix.cpp:661` `pselect` returns 0 -> `break`) with no error surfaced to Dart.
- Expose `flushOutput()` (`tcflush(TCOFLUSH)`) and call it before `close()` to avoid the 30s kernel wait entirely.
- Fix the `tinycthread.c` / glibc >= 2.42 build break.


M0Rf30:
Addressed both review comments in 9a90d1f.

**Double free (`dispose()` -> `closePort()` then `disconnect()` -> helper)** — real bug, fixed. `dispose()` now takes ownership of the `FlSerial` (nulls `_serial`) and clears `serialReadBuff`/`serialWriteBuff` after the synchronous `closePort()`, since flserial frees those buffers but leaves the pointer fields dangling. `_closePortOffUiIsolate()` also returns early when `flh` is already `-1`, so it cannot repeat the cleanup.

**Teardown race with `dispose()`** — `dispose()` now skips its synchronous close when a close is already in flight (tracked in `_pendingNativeClose`), and there is nothing left for it to do in that window: `_fl_close_slot()` sets `breakThread = 1` and `thrd_join()`s the SerialThread *before* it reaches `serialport->close()`. The thread exits immediately (with the all-zero `Timeout()` on Linux/macOS its `read()` returns at once), so the `NativeCallable` can no longer be invoked while the helper isolate waits out the kernel's `closing_wait`.

**Suppressed comment (`Isolate.run` failure)** — also handled: if the spawn fails, the port is now closed inline via `bindings.fl_close(flh)` rather than leaving a native slot and SerialThread alive with no reachable handle. That blocks, but only in the already-degraded path.

M0Rf30:
Second round addressed in b0b5016 — all four points were valid.

**Reconnect race with `fl_free()` (critical)** — `connect()` now awaits `_pendingNativeClose` before the global reset. `fl_close()` in the helper and `fl_free()`/`fl_init()` mutate the same `flserial_tab`, so a reconnect during `closing_wait` could double-close the old slot or kill the newly opened one.

**`dispose()` re-entering `disconnect()`** — when a close is in flight, `dispose()` no longer calls `disconnect()` at all; the running disconnect owns the ordering (native close first, subscription cancel after), so `dispose()` only chains `_closeFrameController` onto the pending close.

**Buffer leak on the non-open dispose path** — buffer release moved into `_releaseSerialBuffers()`, called on every dispose path rather than only after a successful `closePort()`. It frees when the pointers are still live and clears them so a second release is a no-op, which also keeps the double-free fix from the previous round.

**Stale callback after `flh = -1`** — `_handleSerialData()` now returns early unless the status is `connected`. A callback queued before the handle was invalidated made `readList()` throw through `_checkFLH`, publishing a spurious transport error that triggered another disconnect mid-teardown. (No data can be lost at connect time: the subscription is attached and `_status` set to `connected` in the same synchronous block, with no await between.)

`flutter analyze lib/services/usb_serial_service_native.dart` -> no issues.

M0Rf30:
Third round addressed in 10f557e.

**Reconnect vs. the tail of the old teardown (critical)** — correct, `_pendingNativeClose` covered only the native close. `disconnect()` is now a re-entrant-safe wrapper around `_disconnectInternal()`, and `_activeDisconnect` holds the future for the *entire* teardown (native close, subscription cancel, `_status = disconnected`). `connect()` and `dispose()` wait on that, so a reconnect can no longer complete and then get marked disconnected by the old teardown. As a side effect, re-entrant `disconnect()` calls (e.g. `onDone` firing mid-teardown) now join the in-flight one instead of racing it.

**Double free after the synchronous `closePort()`** — correct, and it was mine. `_releaseSerialBuffers()` now takes `alreadyFreed`; after a successful `closePort()` the fields are only zeroed, and the buffers are freed only on the path where `closePort()` did not run.

**Orphaned slot when `isOpen()` is not `open`** — `dispose()` now closes on any live handle (`serial.flh >= 0`) rather than only on `open`, so an I/O-error slot and its SerialThread are no longer left behind. `setDTR()` is in its own try, so a line-control failure cannot skip the close.

**Hot restart while a close is in flight** — not restructured, by design. Hot restart does not call `dispose()` at all; that is exactly why `connect()` opens with the `fl_free()`/`fl_init()` hot-restart guard, which reclaims any thread left over from a dead isolate. Making `dispose()` block until the helper joins would reintroduce the 30s UI freeze this PR exists to remove.

Verification: `flutter analyze` clean, and `flutter build linux --debug` now succeeds — the build break reported in the PR description is the unrelated glibc >= 2.42 / C23 `once_flag` clash in flserial's vendored `tinycthread.h` (it `#define`s `once_flag`/`call_once` unconditionally, mangling glibc's own declarations from `<stdlib.h>`). Patched locally in the checkout; still worth fixing in the fork.

M0Rf30:
Fourth round addressed in fb7958a.

**Android connect not covered by the wait** — valid. The `_activeDisconnect` wait now sits before `_status = connecting`, above the platform branch, so the Android USB-host path is serialized too.

**Connector-level lifecycle not serialized** — valid, and the more important one. `MeshCoreConnector.disconnect()` is now a re-entrant-safe wrapper around `_disconnectInternal()` that publishes the teardown as `_activeDisconnect`; `connect()`, `connectUsb()` and `connectTcp()` join it before starting. Without that, the connector's own teardown tail could still set `_state = disconnected` after a new connection was already up — the transport-level guard could not see it. Repeat `disconnect()` calls now join the in-flight teardown instead of returning early while it is still running (the old `if (_state == disconnecting) return;` returned a completed future while teardown continued).

**`dispose()` returning while a close is in flight (marked critical)** — still by design, same reasoning as the previous round. Hot restart does not call `dispose()`, which is precisely why `connect()` opens with the `fl_free()`/`fl_init()` guard that reclaims threads orphaned by a dead isolate. Blocking `dispose()` until the helper joins would reinstate the 30s UI freeze this PR removes; there is no third option, since the wait is a kernel `close(2)` on a tty the peer is not draining.

Verification on this commit:
- `flutter analyze` -> no issues.
- `flutter test` -> 570 passing; the 2 failures in `message_url_image_helper_test.dart` are pre-existing (identical on a stashed tree).
- `flutter build linux --debug` -> builds.

M0Rf30:
Fifth round addressed in fa9e9d1.

**Connector disposed while a teardown is in flight** — valid. `MeshCoreConnector.dispose()` now sets `_disposed`, and `_setState()`, `notifyListeners()` and `markNotifyDirty()` no-op afterwards, so a teardown that outlives disposal (the off-isolate native close can take the full `closing_wait`) can no longer publish state or arm a debounce timer on a disposed `ChangeNotifier`. The services `_disconnectInternal()` touches on that tail (`_translationService`, `_imageCodecService`, `_imageTransport`) are owned by `main.dart`, not by the connector, and are not disposed here.

**`connecting`-window data dropped** — fair, narrowed. The guard now only rejects the teardown states (`disconnecting`/`disconnected`) rather than everything that is not `connected`, so bytes delivered between the subscription being attached and the status flip are ingested and buffered by the frame decoder. (That window is still zero-width in practice: `_dataSubscription = ...listen(...)` and `_status = connected` sit in the same synchronous block with no await between them.)

**`dispose()` returning during an in-flight close** — unchanged, for the third time by design. Hot restart does not call `dispose()`; the `fl_free()`/`fl_init()` guard at the top of `connect()` is what reclaims a thread orphaned by a dead isolate. Blocking `dispose()` on the helper would restore the 30s freeze this PR removes.

Verification on this commit:
- `flutter analyze` -> no issues.
- `flutter test` -> 570 passing; same 2 pre-existing `message_url_image_helper_test.dart` failures.
- `flutter build linux --debug` -> builds.

M0Rf30:
Sixth round addressed in 9448a31.

**Manual disconnect swallowed by the join** — valid, and a regression I introduced with the re-entrant wrapper. `disconnect(manual: true)` that joins an in-flight automatic teardown now still sets `_manualDisconnect = true`, cancels the reconnect timer and stops the background service before returning the pending future, so the joined teardown can no longer schedule an auto-reconnect the user did not ask for.

**Auto-reconnect after dispose** — valid. `_shouldAutoReconnect` now includes `!_disposed`, which covers both `_scheduleReconnect()` and its timer callback (both consult it), so a teardown outliving disposal cannot arm a timer that calls `connect()` on a disposed connector.

**No re-validation after the await** — valid. `connect()`, `connectUsb()`, `connectTcp()` and `UsbSerialService.connect()` re-check the connection state after awaiting the in-flight teardown; the await is a suspension point, so two callers could pass the initial guard, resume together and both run the open sequence. The re-check and the `connecting` claim are in the same synchronous block, so the claim is atomic with respect to the event loop.

**`dispose()` synchronous-shutdown guarantee** — unchanged; same answer as the previous three rounds. Hot restart does not call `dispose()`, the `fl_free()`/`fl_init()` guard at the top of `connect()` reclaims threads orphaned by a dead isolate, and blocking `dispose()` on a kernel `close(2)` that the peer is not draining is exactly the 30s freeze this PR removes.

Verification on this commit:
- `flutter analyze` -> no issues.
- `flutter test` -> 570 passing; same 2 pre-existing `message_url_image_helper_test.dart` failures.
- `flutter build linux --debug` -> builds.

PR #558, opened 2026-09-13
Save GPX export via file dialog on desktop

M0Rf30:
## Problem

GPX export never works on Linux (and on Windows builds older than 10.0.17763).

`share_plus` cannot share *files* on desktop. `share_plus_linux.dart:22`:

```dart
if (params.files?.isNotEmpty == true) {
  throw UnimplementedError('Sharing files not supported on Linux');
}
```

`GpxExport.exportGPX` always routed through `SharePlus.instance.share(... files: [XFile(path)])`, and the throw was swallowed by the catch-all, so every desktop export ended as `gpxExportFailed`. The already-written `.gpx` was also left behind in the documents directory, because `file.delete()` sits after the throwing call.

Reproduced with a standalone Flutter Linux app:

```
documents dir: /home/gianluca/Documenti
share threw: UnimplementedError: UnimplementedError: Sharing files not supported on Linux
```

## Change

- On `PlatformInfo.isDesktop`, write the file where the user picks via `file_selector`'s `getSaveLocation()` (suggested name `meshcore_<kind>_<utc-timestamp>.gpx`, `gpx` type group). A dismissed dialog maps to `gpxExportCancelled`.
- Mobile keeps the existing share-sheet flow; no temp file is written on desktop any more, so nothing is left behind.
- `file_selector: ^1.0.3` promoted to a direct dependency (already transitive).

## Verification

Standalone Linux probe app: `getSaveLocation()` opened the GTK dialog and returned the chosen path, `RESULT: wrote 20 bytes to /tmp/shareprobe/meshcore_repeaters_probe.gpx`.

`flutter analyze lib/utils/gpx_export.dart` -> no issues.

No in-app run: `flutter build linux` fails in my environment before Dart compilation, in flserial's vendored `tinycthread.c` vs glibc >= 2.42 (`once_flag` typedef redefinition / conflicting `pthread_once`). Pre-existing, unrelated to this change.


zjs81:
Looks good thank you

